PR #33, opened 2024-01-16
zlib: Switch to zlib-ng 2.1.6

joebonrichie:
See https://github.com/snekpit/main/pull/101 for details


der-eismann:
Could you adjust the update ID in the monitoring.yaml to this one please?
https://release-monitoring.org/project/115592/

joebonrichie:
w.r.t. security cpe, zlib-ng doesn't seem to have one yet but it _may_ be affected by any CVEs found in canonical zlib, so idk.

der-eismann (approved):
And this really doesn't even need a rebuild? Fancy.

What bothers me is the package name, as it's "wrong" now, but I guess we're missing the tools to deprecate or conflict packages.

joebonrichie:
Upstream warns there may be subtle ABI differences from different toolchains/flags but yeah it should be drop-in. Safety rebuilds are an option of course.

joebonrichie:
Both fedora, solus and openmadriva have switched now without drama, should be good. Maybe we want to use this pkg as a test harness for moss replaces support?

ikeycode:
Running locally and works grand.

PR #62, opened 2024-01-19
[NFC] Add workflow to detect conflicts

### LINKED ISSUE #5: Add a Github action to check unwanted files in PRs
Just like commit hooks for those with commit privileges, PRs should be guarded by Github actions. PRs shall not include binary files (like .stone files) or other left over files unrelated to packages.

Originally reported by @livingsilver94 

> @der-eismann 
> Hmm, first idea would be something like `for each entry in manifest.jsonc | jq .packages[].files` do a `grep` over the repo to look for duplicates. But I'm sure there is a more elegant solution 😄

der-eismann:
I'd only merge this after fixing the first set of conflicts and/or adjusting the action to only alert for conflicts in modified files. But still, WDYT? Any suggestions regarding behavior or formatting? 
Actual code can be found [here](https://github.com/serpent-os/check-for-file-conflicts/blob/main/src/index.ts)

Fixes #5

### DIFF
--- a/.github/workflows/conflicts.yaml
+++ b/.github/workflows/conflicts.yaml
@@ -1,0 +1,33 @@
+name: Check packages for file conflicts
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  detect_conflicts:
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


ermo (approved):
Let's get this in.

PR #64, opened 2024-01-20
atk: Retire package in favor of `at-spi2-core`

der-eismann:
Turns out that `at-spi2-core` includes all of the files of the `atk` package. Fedora [did the same already](https://src.fedoraproject.org/rpms/atk/blob/255e2ebdd5dd701c23934246aec1c59ceb6d1677/f/dead.package). 
Need to remove the `atk` package manually afterwards as we don't have any `retire` functionality yet.

(see #62)


ikeycode:
Herein lies the issue, `vessel` currently has no method for which to *remove* a package..

der-eismann:
Closing in favor of #123

PR #71, opened 2024-01-20
vala: Update to 0.56.14

der-eismann:
```
root@sosroot # ~ vala --version
Vala 0.56.14
```


ikeycode (approved):
Must continue to support this overwhelmingly popular language ofc

PR #79, opened 2024-01-20
strace: Update to 6.6

der-eismann:
Had to use gcc because there were tons of
```
clang-16: error: cannot specify -o when generating multiple output files
clang-16: error: cannot specify -o when generating multiple output files
clang-16: error: cannot specify -o when generating multiple output files
clang-16: error: cannot specify -o when generating multiple output files
```

```
root@sosroot # ~ strace --version
strace -- version 6.6
Copyright (c) 1991-2023 The strace developers <https://strace.io>.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Optional features enabled: stack-trace=libdw m32-mpers mx32-mpers

root@sosroot # ~ strace echo "hello"
execve("/bin/echo", ["echo", "hello"], 0x7ffe7e756768 /* 21 vars */) = 0
brk(NULL)                               = 0x1116000
arch_prctl(0x3001 /* ARCH_??? */, 0x7fffd89216b0) = -1 EINVAL (Invalid argument)
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/var/cache/ldconfig/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
newfstatat(3, "", {st_mode=S_IFREG|0644, st_size=4071, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 4071, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f8f85fd9000
close(3)
…
```


ikeycode (approved):
Yeah this is one of those packages that *looks* like its using LLVM when it isn't.

PR #86, opened 2024-01-20
wlroots: Add at 0.16.2

der-eismann:
Also adding `seatd` as a dependency. Will be used for sway.


ikeycode (approved):
Let the gooey testing begin

PR #118, opened 2024-01-26
fontconfig: Add trigger to update fonts cache

der-eismann:


ikeycode:
We wanna bump relno? :D

ikeycode (changes requested):
need relno bump

ikeycode (approved):
ty ty ty

PR #122, opened 2024-01-27
Fix texinfo trigger & remove conflicting `/usr/share/info/dir` files

der-eismann:


ermo (approved):
Why not?

der-eismann:
Exactly! :smile:

PR #123, opened 2024-01-27
atk: Make package empty to resolve conflicts

der-eismann:
I just noticed that this doesn't produce any stone files... probably won't work as well, right? Do we need an empty stone?


ikeycode:
We'll need to match the original patterns - stubbing each subpackage with empty/doc files ?

ikeycode:
Superseded by 52abafa083f5c17c3bb516040c3babc53855b805

PR #125, opened 2024-01-28
Enhance sway ootb experience 

der-eismann:
Still not perfect, but it's getting better

screenshot-20240128-181659


ikeycode (approved):
lgtm

PR #130, opened 2024-01-30
Add libxmlb & appstream

der-eismann:
Stumbled over them a few times in the GNOME universe, might as well add them when I have already built them.


a/appstream/stone.yml
```diff
@@ -0,0 +1,45 @@
...
+    - pkgconfig(libcurl)
+    - pkgconfig(libsystemd)
+    - pkgconfig(libxml-2.0)
+    - pkgconfig(libzstd)
+    - pkgconfig(xmlb)
+    - pkgconfig(yaml-0.1)
+toolchain   : gnu
+setup       : |
+    sed -i '/missing-include-dirs/d' meson.build
+    %meson -Dstemming=false -Dvapi=true -Dsvg-support=false -Dapidocs=false -Dinstall-docs=false --sysconfdir=/usr/share/appstream
```
ermo, line 41: Please use the %(datadir) definition (defined here: https://github.com/serpent-os/boulder/blob/main/data/macros/arch/base.yml#L11)

ermo (changes requested):
Small nit, otherwise looks good.

a/appstream/stone.yml
der-eismann: Fixed!

ermo (approved):
LGTM, thanks!

PR #138, opened 2024-02-02
appstream: Enable svg support

joebonrichie:


a/appstream/stone.yml
```diff
@@ -17,19 +17,18 @@ description : |
...
     - docbook
     - binary(cmake)
     - binary(gperf)
     - binary(itstool)
     - binary(msgfmt)
     - binary(vapigen)
     - binary(xsltproc)
-    - pkgconfig(gio-2.0)
     - pkgconfig(glib-2.0)
-    - pkgconfig(gobject-2.0)
```
der-eismann, line 30: Why the removal?

der-eismann (approved):
Thanks!

ikeycode:
I get the concern re explicit build deps but I can't fault anyone until we have the right tooling + documentation.
As for glib-2.0 gio-2.0 gio-unix-2.0 - currently they're all in the same package so that's not an issue yet :)

PR #139, opened 2024-02-02
libadwaita: Initial inclusion at 1.4.2

der-eismann:


ikeycode (approved):
Note to self: Add jankiness test for whole repo to determine proportion of GNU builds..

PR #140, opened 2024-02-05
Add gnome text editor

der-eismann:


ikeycode (approved):
note to self: `update-desktop-database` needs sorting + triggerfying

nice work!

PR #152, opened 2024-02-15
zellij: Add at 0.39.2

tarkah:


z/zellij/monitoring.yml
```diff
@@ -0,0 +1,2 @@
+releases:
+  id:
```
der-eismann, line 2: Drop a [300087](https://release-monitoring.org/project/300087/) here :slightly_smiling_face:

tarkah: Done, what does this relate to?

ermo (changes requested):
Please use "zellij: Add at <version>" in the header. Makes it easier to follow monorepo activity in git logs.

We also traditionally put the summary key just above the description key for easy reference.

ermo (approved):
Optional: 

Play with adding e.g. `--release --config debug="full"` 

(per https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo and `man cargo`+search for `--config`)

I know you need zellij to feel productive so it's pointless to hold this up for any longer.

PR #153, opened 2024-02-16
helix: Add at 23.10

tarkah:


h/helix/pkg/remove-grammars.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/languages.toml b/languages.toml
+index 77c629e4..f4ca3b03 100644
+--- a/languages.toml
++++ b/languages.toml
+@@ -1,6 +1,9 @@
+ # Language support configuration.
+ # See the languages documentation: https://docs.helix-editor.com/master/languages.html
+ 
++# Tree-sitter sources 403 or build fails (d, rescript)
++use-grammars = { except = [ "hare", "wren", "gemini", "d", "rescript" ] }
```
tarkah, line 19: As if not a sign from the gods, the `d` tree-sitter grammar fails to compile in boulder with:

`error: call to undeclared function 'isascii'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]`

I'll re-enable if anyone knows how to fix this

ermo (approved):
LGTM!

PR #154, opened 2024-02-17
starship: Add at 1.17.1

tarkah:


ermo (changes requested):
@tarkah, can we make this the smoke-test for the new cargo macros?

tarkah:
> @tarkah, can we make this the smoke-test for the new cargo macros?

Done!

PR #156, opened 2024-03-06
Add desktop-file-utils recipe

flippedbits:
I have added this because it seems to come up as a requirement when building other packages often and seemed like a good first packaging attempt.


d/desktop-file-utils/stone.yml
```diff
@@ -0,0 +1,24 @@
+#
+# SPDX-FileCopyrightText: © 2020-2024 Serpent OS Developers
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : desktop-file-utils
+version     : 0.27
+release     : 1
+homepage    : https://www.freedesktop.org/software/desktop-file-utils/releases
```
ikeycode, line 9: I think this needs to be https://www.freedesktop.org/wiki/Software/desktop-file-utils/ - sorry!

d/desktop-file-utils/stone.yml
```diff
@@ -0,0 +1,24 @@
...
+license     :
+    - GPL-2.0-or-later
+builddeps   :
+    - pkgconfig(glib-2.0)
+setup       : |
+    %meson
+build       : |
+    %meson_build
+install     : |
+    %meson_install
```
der-eismann, line 24: A trailing newline is always welcome here, most of the time you can set this in your editor of choice :slightly_smiling_face:

ikeycode (approved):
Ty! Will merge squashed

PR #163, opened 2024-03-07
perl-module-build: Add at 0.4234

joebonrichie:
Prerequisite to build modern perl modules.

Once this is landed we can land perl macros in boulder

Depends on #164


joebonrichie:
Changing to draft as our perl paths have the full version in the path meaning we'll have to do rebuilds for every point release i.e. `/usr/lib/perl5/vendor_perl/5.38.2`

joebonrichie:
Module::Build is deprecated, closing. We'll rely on ExtUtils::MakeMaker instead.

PR #171, opened 2024-03-09
perl: Improve build portability

joebonrichie:
Don't hardcode x86_64. Should help with aarch64 in future.


ikeycode (approved):
Thank you for considering portability <3

PR #172, opened 2024-03-09
Samba

joebonrichie:


ikeycode (changes requested):
I think for `ldb` we need an extra configure:

```
--with-modulesdir=%(libdir)/ldb/modules
```

?

ikeycode (approved):
much appreciated ty!

PR #174, opened 2024-03-10
Gtk4 vulkan

joebonrichie:
Packages using gtk-4 will need to be rebuilt to make use of it.


ikeycode (approved):
thank *gawd* - shaderc was giving me headaches!

PR #179, opened 2024-03-10
gnome-control-center: Add rundep on tecla

joebonrichie:
- To visually view keyboard layout


ikeycode (approved):
good catch, thank you!

PR #181, opened 2024-03-11
Ca certs

joebonrichie:
Be aware this won't build with boulder-rs due to differences in directory layout.


ermo:
@joebonrichie Now that boulder-d has been retired and boulder-rs has become the default, is there any chance you could find the time to update this maybe...?

ikeycode:
Cherry-picked and merged into `main`

PR #192, opened 2024-04-22
Autobuild rebuilds

ermo:
When using `autobuild query src:. systemd`, the recipes in this PR showed up as unreadable during the tree scan that `autobuild` performs at startup when creating its DAG.

Additionally, boulder has grown the ability to emit build-release IDs in .jsonc manifests, so as an experiment, I've rebuilt the packages that the `autobuild` run flagged as non-compliant with the OG DLang libmoss binary moss-format using `boulder build -b2` for each.


ermo:
Converting this to draft, pending outcome of https://github.com/serpent-os/moss/issues/210

ermo:
Bumped to `release: 3` to ensure that it replaces existing python-eopkg (currently `release: 2`) if it gets merged.

PR #196, opened 2024-07-05
Tomoyo tools for linux-kvm 6.9.8

ermo:
- Add tomoyo_tools at v2.6.1
- Update kernel to 6.9.8 and integrated TOMOYO config variables with tomoyo-tools and systemd-as-PID1

The goal with this PR is to be able to begin experimenting with whether TOMOYO can be used to "seal" access to `/usr` and `/.moss` on Serpent OS systems for now or until we find a better solution.


ermo:
Currently testing this some more; doesn't appear to be able to mount the root fs atm.

ikeycode (approved):
LGTM - we will need a subsequent build to port to new blsforme layout though, check linux-desktop for reference =)

PR #199, opened 2024-07-20
dracut: Update to v103 and add patches

ReillyBrogan:
Switch to the dracut-ng fork which is maintained. Add several patches of note:
- Invalidate depmod index module. Used for initrd extension
- Skip firmware installation in initrd. Used for linux-firmware initrd extension
- Skip hwcaps directories when processing libs. Ensures that built initrds are usable on machines with a lower ISA than supported by hwcaps
- Stateless ldconfig configuration
- De-compress kbd files prior to initrd creation. Improves initrd compression ratio


d/dracut/stone.yaml
```diff
@@ -24,6 +24,12 @@ rundeps     :
     - binary(cpio)
     - terminus-font
 setup       : |
+    %patch %(pkgdir)/patches/downstream/0001-Add-module-to-invalidate-depmod-index.patch
```
ikeycode, line 27: This seems Solus specific not sure where/how it integrates into Serpent atm. Any upstream tracking bug for this in dracut-ng?

d/dracut/stone.yaml
```diff
@@ -24,6 +24,12 @@ rundeps     :
...
+    %patch %(pkgdir)/patches/downstream/0001-Support-stateless-glibc-ldconfig-configuration.patch
+    %patch %(pkgdir)/patches/downstream/0003-Exclude-unwanted-paths-from-initrd-creation.patch
+    %patch %(pkgdir)/patches/downstream/0004-Downstream-De-compress-kbd-files.patch
+    %patch %(pkgdir)/patches/downstream/0001-Allow-for-skipping-firmware-installation.patch
```
ikeycode, line 31: Also seems Solus specific, would be nice to get an upstream bug in dracut-ng to track progress

d/dracut/stone.yaml
ReillyBrogan: This provides the functionality that we talked about where depmod is ran in the initrd when an initrd extension is present with kernel modules. Without this the addon kernel modules will not be available which defeats the purpose of having initrd modules in the first place.

d/dracut/stone.yaml
ReillyBrogan: Not really a point tbh I'm not sure of any distro that is taking this approach with the initrd.

d/dracut/pkg/patches/downstream/0003-Exclude-unwanted-paths-from-initrd-creation.patch
```diff
@@ -0,0 +1,114 @@
...
+ # Version comparison function.  Assumes Linux style version scheme.
+diff --git a/src/install/dracut-install.c b/src/install/dracut-install.c
+index 7473e87a..4ab8ea52 100644
+--- a/src/install/dracut-install.c
++++ b/src/install/dracut-install.c
+@@ -757,8 +757,39 @@ static int dracut_mkdir(const char *src)
+         return 0;
+ }
+ 
++const char *strip_avx_path(const char *fullpath, const char *avxpath)
```
ikeycode, line 48: I'm quite nervous about this patch in particular. Note `strncat` is preferred and errors checked.

Also note `strstr` returns a pointer within existing memory whether allocated or not. This one is returning existing memory *or* newly allocated memory which seems like its going to explode or cause leaks.

ikeycode:
ive rebased these for you, for now the exclude paths patch is disabled as my feedback was never replied to, and i have concerns with it. i also didn't want to waste your efforts on uplifting the package

PR #200, opened 2024-07-20
linux-headers: Update to v6.10

ReillyBrogan:


ikeycode (commented):
I'm actually not sure how I feel about this. Keep in mind we had to roll our glibc back to older kernel baseline to ensure compatibility with older kernels on builders not running Serpent.

Following mainline seems a bit excitable ?

ReillyBrogan:
> I'm actually not sure how I feel about this. Keep in mind we had to roll our glibc back to older kernel baseline to ensure compatibility with older kernels on builders not running Serpent.
> 
> Following mainline seems a bit excitable ?

Linux-headers version and glibc min kernel version are independent, with the exception that glibc min kernel baseline should always be lower than the linux-headers version. Updating linux-headers to v6.10 just means that when glibc is next rebuilt that it be able to use kernel syscalls/apis that are more optimal when booted on a newer kernel. It will still use appropriate syscalls/apis when booted on a kernel older than 6.10 (down to whatever the glibc kernel baseline is). 

Keeping linux-headers at 6.6 means that glibc will only ever use syscalls/apis that are optimal for 6.6 when booted on a kernel >= 6.6. That isn't really ideal.

TLDR updating linux-headers doesn't change the minimum kernel version, it just means that glibc is able to utilize newer kernels better.

ikeycode (approved):
Let's give it a go. Scheduled builds from a phone, magic.

PR #206, opened 2024-07-27
niri: add at 0.1.7

jpeeler:
An extremely basic recipe. Maybe it's useful as a starting point? It does at least work in Qemu:

screenshot


ReillyBrogan:
You need to add all of the other files in the directory that are created once the package is built.

ReillyBrogan (changes requested):
See other comment

ikeycode (approved):
Lgtm, takes advantage of the cosmic tricks I see :D

PR #207, opened 2024-08-01
GCC 14 and cleanups

ReillyBrogan:
See gcc commit for full changes.

I was initially planning on cleaning up glibc and binutils as well, but I think I'll save that for a future PR. Note that once this is done being built that each of these packages will need to be rebuilt again to complete bootstrapping.

I'll be opening up a PR for boulder to update the list of GNU toolchain packages to install. This should build without that though, but packages might need a manual builddep on g++ until that is complete.

Also note that we'll probably need to rebootstrap GCC entirely to add Ada support since that needs an existing GNAT toolchain to build. Since we need to bootstrap ARM at some point anyway hopefully we won't need it until then.


ikeycode (approved):
lets go

PR #211, opened 2024-08-02
ncurses: Fix pkgconfig

ReillyBrogan:
Ncurses was leaking the LDFlags into the pkgconfig file. Boulder adds `--icf=safe` by default when building with LLVM, which was getting put into the flags in the pkgconfig file. Since binutils doesn't support this flag that caused any gnu toolchain builds that tried to link against ncurses to fail.

Also, switch the C++ bindings to be a shared library


ikeycode (commented):
what happened to all the terminfo files..?

ReillyBrogan:
> what happened to all the terminfo files..?

`/usr/lib/terminfo` is now a symlink to `/usr/share/terminfo`. The files were previously duplicated.

ikeycode:
Sweet :)

ikeycode (approved):

PR #212, opened 2024-08-02
flatpak: Enable flathub by default

joebonrichie:


joebonrichie:
the other option for handling this is patch flatpak for statelessness support and dump the `.flatpakrepo` file into the remotes.d directory.

ikeycode (approved):
Decent approach but eventually I'd like if we upstream vendored repo support into flatpak

PR #217, opened 2024-08-03
fix niri, add missing files

jpeeler:
Nothing but the niri binary was included before. Includes the wayland and systemd configuration files.


ikeycode (changes requested):
Don't forget we need the manifest files:)

jpeeler:
Argh, sorry about that. Don't have a good dev workflow to make this more obvious currently.

ikeycode (approved):
Lgtm :)

PR #218, opened 2024-08-03
udisks: Enable daemon & statelessness support

joebonrichie:


l/libbytesize/stone.yaml
```diff
@@ -0,0 +1,33 @@
+#
+# SPDX-FileCopyrightText: © 2020-2024 Serpent OS Developers
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : libbytesize
+version     : 2.10
```
ermo, line 7: version literal needs to be quoted; otherwise YAML will interpret it as `2.1` (see associated jsonc manifest `source-release` value above).

ermo (changes requested):
Apart from a small nit, LGTM.

I successfully built and tested this from a local repo. G-C-C shows disk size with this PR installed.

ikeycode (approved):

PR #230, opened 2024-08-05
tools: Add prepare-commit-msg hook

joebonrichie:
This will try to prefill the commit msg based on the diff, e.g. pkg: Update to n.n.n or pkg: Add at n.n.n

It'll also give some guidance for writing the commit message.

Use `just init` to setup.

### DIFF
--- a/justfile
+++ b/justfile
@@ -31,3 +31,8 @@ _clean target:
 
 # Clean *.stone artefacts from the current directory
 clean: (_clean build_file)
+
+# Init git hooks
+_init target:
+    ln -sfv ../../tools/prepare-commit-msg.py $(git rev-parse --git-path hooks)/prepare-commit-msg
+init: (_init build_file)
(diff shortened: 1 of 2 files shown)


joebonrichie:
Will need some handling here whether pyyaml is installed or not, otherwise feel free to try out

ermo:
> Will need some handling here whether pyyaml is installed or not, otherwise feel free to try out

Currently, pyyaml is not available in the repos, so the smart move is probably to package it up as part of this PR...?

joebonrichie:
uses json from stdlib to parse the manifest.jsonc file instead to avoid dependencies. JSONDecoder is overridden to handle the comment in the .jsonc file.

joebonrichie:
Automatic "Add at ${version}" hooked up

ermo (approved):
I like this.

PR #234, opened 2024-08-07
rustls-ffi: drop clang-asan.patch

cpu:
Hi there, I was interested to learn that you're shipping `curl` with `rustls-ffi` by default. Very cool :tada: 

While looking at the package definition I noticed you were carrying an extra patch that I think can be dropped. I'm not at all familiar with SerpentOS so hopefully I haven't made any silly errors in this contribution.

Previously the `rustls-ffi` build setup applied a patch to the upstream `Makefile` to remove unconditional use of ASAN w/ `clang`.  It looks like it was added [alongside the original addition](https://github.com/serpent-os/recipes/commit/0b35098158f7d379d4c2f35cfb08fe5e7343d1c5) of `rustls-ffi`. This branch removes the patch because it's no longer necessary for two reasons:

1. The package was [already migrated](https://github.com/serpent-os/recipes/commit/b45b010f1ed12a91d2319a1246316e42f9d109d7) to build using `cargo-c` and so I believe hasn't been using the patched `Makefile` since then.
2. We independently noticed & fixed [this bug](https://github.com/rustls/rustls-ffi/issues/423) upstream ([the fix](https://github.com/rustls/rustls-ffi/pull/425) will be included in the next release). The intent was to only enable ASAN for debug builds, not release builds (the `Makefile` default). Sorry about that!


ikeycode (approved):
Hey @cpu thanks for the PR and for rustls-ffi! Anywhere we can displace insecure surface area in Serpent OS with libs like this, we will.

I'm not gonna get picky with the PR because honestly our docs are crap and we're still in "open the doors up" mode, so I'll do the relevant bump and build atop this using `boulder` (from the `moss` repo)

Thanks! =)

cpu:
Thanks! Don't be shy if you run into any issues w/ rustls-ffi in the future. Happy to help.

PR #238, opened 2024-08-08
niri: fix misspelling of systemd unit config file

jpeeler:
Sorry for the noise... I tested this now. Should be no more issues.

(Thought about waiting to fix at the next release, but sounds like it's going to be a while before that occurs based on discussion in the niri room.)


ikeycode (approved):
get it in. bumps are "free" :D

PR #249, opened 2024-08-10
upower: Update to 1.90.4 & make stateless

joebonrichie:
[Changelog](https://gitlab.freedesktop.org/upower/upower/-/raw/v1.90.4/NEWS?ref_type=tags)


joebonrichie:
this is an important one to land as it fixes high battery drain

ikeycode (approved):
much appreciated

PR #268, opened 2024-08-12
[NFC] Fix double quotes in example `.env` file content

der-eismann:
This just cost me 10 minutes to figure out :smile:


ermo (approved):
My bad. Thanks for fixing it!

PR #269, opened 2024-08-13
Python updates

joebonrichie:
Depends on #359


joebonrichie:
`python-pep517` will need manual removing from index after this lands

ikeycode (approved):

PR #272, opened 2024-08-14
portal: Add at v1.2.3

Nivalux:
My first recipe, please double check :)
I use portal to copy files between my PC, VM and Laptop.


ermo:
For now, this is out of scope for us. You are of course welcome to maintain it in a local repo until we get our ducks in a row for the rebuild story in the official repo.

Until we do, we are striving to only include what we really need for minimal functionality for our emerging workflows.

ikeycode:
Unless i'm mistaken this is being added by @Nivalux so they can better contribute? its a file sending thingy

Nivalux:
> Unless i'm mistaken this is being added by @Nivalux so they can better contribute? its a file sending thingy

It is, but I can also use it from my personal repo. 
So if no one else needs it, it can remain there for now :)

PR #274, opened 2024-08-14
lmdb: Autogenerate correct lmdb.pc on each build

ermo:


ReillyBrogan:
This seems unnecessarily complicated compared to just having `lmdb.pc.in` in the pkg folder and using sed in the build script to set the version.

ReillyBrogan:
It's still more complicated than just doing the following:
```
    sed -e "s|@PKGVER@|%(version)|g" %(pkgdir)/lmdb.pc.in > lmdb.pc
    %install_file lmdb.pc %(installroot)/%(libdir)/pkgconfig/lmdb.pc
```

ermo:
It's also way more fun.

ikeycode (approved):

PR #275, opened 2024-08-15
golang: Update to 1.22.6

ermo:
Packaging changes:
- GOROOT_FINAL was deprecated for 1.21, so removed that from the build
- CGO_FLAGS hack is no longer necessary, so removed that
- Update stateless patch

Release notes:
- [1.21.0](https://go.dev/doc/go1.21)
- [1.21.x](https://go.dev/doc/devel/release#go1.21.0)
- [1.22.0](https://go.dev/doc/go1.22)
- [1.22.x](https://go.dev/doc/devel/release#go1.22.0)


ReillyBrogan:
I'm not really sure how it's supposed to work it's way into the install but we need `go.env` and a patch like this:
https://github.com/getsolus/packages/blob/main/packages/g/golang/files/0001-solus-Use-local-toolchain.patch

Without this packages might end up trying to download their own golang binaries during the build and ignore the system one.

ikeycode (approved):

PR #280, opened 2024-08-16
niri: update to 0.1.8

jpeeler:
released Aug 10, 2024


n/niri/stone.yaml
```diff
@@ -4,13 +4,13 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 name        : niri
-version     : "0.1.7"
-release     : 3
+version     : "0.1.8"
+release     : 1
```
ikeycode, line 8: Oh - why a relno decrease? We bump this as a monotonic integer =)

jpeeler: Too used to RPM packaging. Also, I assume that a rebuild was necessary to update the binary (does it just mirror the contents of jsonc)?

ikeycode: Yeah binary manifest is proof of build :D

ikeycode (approved):

PR #335, opened 2024-08-20
Slibtool: Update to v0.6.0

ReillyBrogan:
This SUCKED.

PR #336, opened 2024-08-20
Avahi

ReillyBrogan:
Avahi has accumulated numerous fixes (including for CVEs) but hasn't had a release in some time. Use the git source instead.

Note that we probably want the systemd services enabled, but that needs to be coordinated between systemd, networkmanager, and avahi so that is saved for a later time.

Depends on serpent-os/recipes#335


ikeycode (approved):

ikeycode:
I think using presets should be enough for preenable right?

ReillyBrogan:
Probably, it's more a question of making sure it's configured correctly so things aren't conflicting

PR #339, opened 2024-08-20
Ncurses/readline/bash

ReillyBrogan:


ikeycode (changes requested):
This restores the broken Solus patch for stateless bash, rather than relying on the profile files to then load `/etc/profile.d/*` if they exist. :)

ikeycode:
closing as joeys PR revamped bash, and i never got any more replies on this PR

PR #361, opened 2024-08-31
fd: Add at v10.2.0

jplatte:
This is a much better `find`, I use it locally somewhat frequently. Just `fd searchterm searchdir`, very similar to ripgrep and uses some of the same underlying crates (e.g. for respective VCS ignore files).

Regarding packaging (this is the first thing I packaged), I copied ripgrep and adjusted. Unclear why there is a `target/{triple}/release` here and a `target/{triple}/debug` for ripgrep (to run the binary from for shell completion generation). Also is there a story for completions for other shells yet? I found a zsh file in `contrib` of the upstream repo.

~~Haven't been able to install this locally, says not found for some reason (it worked for cryptsetup).~~

Project repo for reference: https://github.com/sharkdp/fd


ikeycode (changes requested):
Actually looks great but we'll need a `monitoring.yaml` for https://release-monitoring.org/project/373722/

jplatte:
Is that website related to the Serpent OS project? Doesn't look like it. I have never seen it before, looks nice :)

ikeycode:
> Is that website related to the Serpent OS project? Doesn't look like it. I have never seen it before, looks nice :)

I wish! Part of the fedora landscape, super helpful !

ikeycode (approved):

PR #363, opened 2024-09-02
Various useful recipes/ tree tweaks to git hooks and justfile

ermo:
**Add {create,index,ls,mv}-local targets**

All targets use the LOCAL_REPO environment variable.

The default LOCAL_REPO location is set to:

    ${HOME}/.cache/local_repo/x86_64

for XDG compliance.

This can trivially be overridden in the `recipes/.env` file or via an
environment variable, which is expected to be useful for stack work.

If the LOCAL_REPO dir does not exist, it will be created first. This is
also intended to be useful for stack work.

Note that the options for the ls-local target file listing have been
carefully chosen so the list is sorted by file creation time, which is
useful for checking if things have been built in the correct order and
whether all files present have been indexed.

The README has been updated to reflect the defaults above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ We'll add the (unversioned) volatile repository¹ at the bottom layer, and eleva
 local repository priority to take precedence.
 
 ```bash
-$ boulder profile add local-x86_64 --repo name=volatile,uri=https://dev.serpentos.com/volatile/x86_64/stone.index,priority=0 --repo name=local,uri=file:///${HOME}/.local_repo/stone.index,priority=10
+$ boulder profile add local-x86_64 --repo name=volatile,uri=https://dev.serpentos.com/volatile/x86_64/stone.index,priority=0 --repo name=local,uri=file://${HOME}/.cache/local_repo/x86_64/stone.index,priority=10
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,62 +16,79 @@ $ sudo usermod --add-subuids 1000000-1065535 --add-subgids 1000000-1065535 root
...
-**Create a profile**
+**Create a boulder build profile**
 
-We'll add the (unversioned) volatile repository¹ at the bottom layer, and elevate
-our local repository priority to take precedence.
+We'll add the (unversioned) volatile repository¹ at the bottom layer, and elevate our
+local repository priority to take precedence.
 
 ```bash
 $ boulder profile add local-x86_64 --repo name=volatile,uri=https://dev.serpentos.com/volatile/x86_64/stone.index,priority=0 --repo name=local,uri=file:///${HOME}/.local_repo/stone.index,priority=10
```
jplatte, line 59:
```suggestion
$ boulder profile add local-x86_64 --repo name=volatile,uri=https://dev.serpentos.com/volatile/x86_64/stone.index,priority=0 --repo name=local,uri=file://${HOME}/.cache/local_repo/x86_64/stone.index,priority=10
```

ikeycode (approved):

PR #376, opened 2024-09-12
Issues Template: New templates for hardware enablement and package requests

rav101:
Created a new hardware enablement template for Serpent recipes repo. There is a lot stated as required fields but we can tweak as appropriate


rav101:
Updated to also add a new package request template

ikeycode (approved):

Issue #353, opened 2024-08-25
list of packages to consider adding


kojq:
There are many packages we could bring over, but I am attempting to list a small set of what I find to be relevant packages for long-term experience on desktop (can link pertaining PRs to this issue).

- [ ] bluez
- [ ] booster
- [ ] crun
- [ ] cuda
- [ ] cups-pdf
- [ ] dbus-broker-units
- [ ] distrobox
- [ ] espeak-ng
- [ ] iptables-nft
- [ ] keepassxc
- [ ] libva-mesa-driver
- [ ] nvidia-open
- [ ] logrotate
- [ ] mesa-vdpau
- [ ] noto-fonts
- [ ] noto-fonts-cjk
- [ ] noto-fonts-emoji
- [ ] obs-studio
- [ ] openconnect
- [ ] podman
- [ ] qt6-wayland
- [ ] speech-dispatcher
- [ ] system76-power
- [ ] systemd-resolvconf
- [ ] typst
- [ ] vulkan-intel
- [ ] vulkan-radeon
- [ ] vulkan-nouveau
- [ ] vulkan-validation-layers
- [ ] wgcf
- [ ] zram-generator

Possible in some way:
- [ ] cachy-browser (am using this, though ladybird, zen, verso, etc are growing projects to eye)
- [ ] fcitx5 (IME versus switching locale)
- [ ] linux-cachyos-lto (interested in the patches and other changes, including ThinLTO)
- [ ] sbctl (depending on timing of SerpentOS supporting SB itself, and how that would work with nvidia-open)

A few packages for those interested in emulation:
- [ ] cemu
- [ ] dolphin-emu
- [ ] ryujinx

ikeycode:
Some of these look like distro-specific splitting names, could we cross-reference with our current recipes to ensure nothings already packaged, and make sure we're using upstream names? would be super helpful!

ermo:
Updated the list with the noto fonts I _know_ we have.

Pheoxy:
I've already packaged `system76-power` locally a while ago but cooked the install for serpentos.
I need a bit of free time to be able to get that and some others upstream though sorry so give me a few days.

Seemed to be working but I'm literally just mixing and matching until it builds as I'm not new to building my own packages but also I'm not the most experienced. 😅

kojq:
> Some of these look like distro-specific splitting names, could we cross-reference with our current recipes to ensure nothings already packaged, and make sure we're using upstream names? would be super helpful!

[...]

PR #396, opened 2024-09-25
system76-power: Add at v1.2.1

Pheoxy:
https://github.com/serpent-os/recipes/issues/353


s/system76-power/stone.yaml
```diff
@@ -0,0 +1,27 @@
...
+version     : 1.2.1
+release     : 1
+homepage    : https://github.com/pop-os/system76-power
+upstreams   :
+    - https://github.com/pop-os/system76-power/archive/refs/tags/1.2.1.tar.gz: 90b796e72a442a7968fc80b269ed2a936c6541bcba939a72612898082fbd7ec3
+summary     : system76-power is a utility for managing graphics and power profiles.
+description : |
+    system76-power is a utility for managing graphics and power profiles.
+license:
+    - GPL-3.0
```
joebonrichie, line 16: nit: use a valid spdx licence identifier. Looks like it's `GPL-3.0-only` in this case https://github.com/pop-os/system76-power/blob/master/src/acpi_platform.rs#L2

s/system76-power/stone.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - binary(cargo)
+    - pkgconfig(libusb-1.0)
+rundeps     :
+    - pkgconfig(dbus-1)
+    - pkgconfig(libusb-1.0)
+networking  : yes
+build       : |
+    %make
+install     : |
+    %make sysconfdir="/usr/lib" DESTDIR="%(installroot)" install
```
joebonrichie, line 27: nit: there is a `%make_install` macro

s/system76-power/stone.yaml
Pheoxy: Done and thanks I've added a this to #https://github.com/serpent-os/serpentos.com/issues/20 as well.

s/system76-power/stone.yaml
Pheoxy: Changed to `%make_install`

Pheoxy:
Probably should have done another commit for reviewed changes...

joebonrichie (approved):

PR #402, opened 2024-10-05
gmp: Update to v6.3.0

Pheoxy:
### Update GMP to v6.3.0

This pull request updates the GNU Multiple Precision Arithmetic Library (GMP) with the following enhancements:

#### Enhancements

1. **LTO (Link Time Optimization) Support**:
   - Added LTO support to improve the performance and optimization of the GMP library.

2. **Additional Make Checks**:
   - Included missing `make check` commands to ensure comprehensive testing during the build process.


ReillyBrogan (changes requested):
Commits should be squashed, also we recently(ish) enabled LTO by default. That flag is no longer needed if you want it to continue using thin-lto. Alternately you can switch it to full LTO with `lto: full`

Pheoxy:
> Commits should be squashed, also we recently(ish) enabled LTO by default. That flag is no longer needed if you want it to continue using thin-lto. Alternately you can switch it to full LTO with `lto: full`

Thanks for the update. I deliberately kept the commits separate because I believe package version updates and other changes should remain distinct within the PR. Squashing them would mix these changes, making debugging more difficult later—even for a simple update like this.

Regarding the LTO flag, I appreciate the heads-up.

I don’t have the time to update this PR myself at the moment, but if someone else wants to take over, they’re welcome to.

ReillyBrogan:
Superceded by AerynOS/recipes#683 (I integrated all of your other changes into my version update).

PR #406, opened 2024-10-10
harfbuzz: Update to v10.0.1

joebonrichie:
[Changelog](https://raw.githubusercontent.com/harfbuzz/harfbuzz/refs/tags/10.0.1/NEWS)


h/harfbuzz/stone.yaml
```diff
@@ -4,18 +4,17 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 name        : harfbuzz
-version     : 9.0.0
-release     : 9
+version     : 10.0.1
```
ReillyBrogan, line 7: Looks like 10.4.0 is the latest now

h/harfbuzz/manifest.x86_64.jsonc
```diff
@@ -20,21 +20,20 @@
 				"soname(libgobject-2.0.so.0(x86_64))",
 				"soname(libgraphite2.so.3(x86_64))",
 				"soname(libicuuc.so.75(x86_64))",
-				"soname(libm.so.6(x86_64))",
-				"soname(libstdc++.so.6(x86_64))"
```
ReillyBrogan, line 24: It's very interesting that this disappears but that it didn't link against libc++ instead. I don't think it's an issue, just surprised to see that.

ReillyBrogan (approved):
I guess it doesn't matter that there's a new version out, this is still newer than what we have.

PR #407, opened 2024-10-12
slibtool: simplify symlink

firasuke:


ikeycode (approved):
Thanks! We like simple =)

firasuke:
> Thanks! We like simple =)

Glad I could help!

PR #408, opened 2024-10-13
llvm: Speed up clang by 1.22x

joebonrichie:
Build libLLVMSupport to generate profiling data, previously we were only compiling a very simple HelloWorld program for profiling.

This speeds up building llvm with clang by 1.22x

Additionally build with thin LTO instead of full as the memory requirements for fat LTO are too high with little gain.


ikeycode (approved):
love it

PR #420, opened 2024-11-22
gnome-tweaks: Add at v9999-git (track git master)

ermo:
I need it for being able to conveniently tweak my GNOME VM font sizes and zoom factor for text accessibility settings, and it's no longer available as a flatpak from what I can tell...


ikeycode (approved):

ikeycode:
le ty - and ty for fixing gir for gudev!

PR #440, opened 2024-12-14
tools: Add new function to bash helpers and new helpers for other shells 

malfisya:
**Summary**

- Add `gotoserpentrepo` function to bash helper script
- Add basic script to navigate the monorepo in fish and zsh

### DIFF
--- a/tools/helpers.bash
+++ b/tools/helpers.bash
@@ -73,6 +73,14 @@ function cpesearch() {
     fi
 }
 
+# Goes to the root directory of the serpent recipes
+# git repository from anywhere on the filesystem.
+# This function will only work if this script is sourced
+# by your bash shell.
+function gotoserpentrepo() {
+    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../" || return 1
+}
+
 # Goes to the root directory of the git repository
 function goroot() {
     cd "$(git rev-parse --show-toplevel)" || return 1
--- a/tools/helpers.fish
+++ b/tools/helpers.fish
@@ -1,0 +1,37 @@
+#!/usr/bin/env fish
+function __serpent_repo_dir
+    realpath (dirname (readlink (status -f)))/../
+end
+
+function __serpent_toplevel
+    git rev-parse --show-toplevel
+end
+
+function cpesearch -d "Search for known CPE entries for a program or library"
+    if test "$argv[1]" = "--help"; or test "$argv[1]" = "-h"; or test (count $argv) -ne 1
+        echo "usage: cpesearch <package-name>"
+    else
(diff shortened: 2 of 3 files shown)


malfisya:
I only tested the bash script and the experience wasn't not starightforward. In Solus, I can just symlink the script to `~/.bashrc.d/` and it will just work. In Serpent, I have to copy the `/usr/share/defaults/etc/profile` to `/etc/bashrc` and added  this line (mimiicking what Solus does in its equivalent file):

```
if [ -d ~/.bashrc.d ]; then
    for rc in ~/.bashrc.d/*; do
	if [ -f "$rc" ]; then
	    . "$rc"
	fi
    done
    unset rc
fi
```
I have not tested  the zsh and fish script. Cheers!

ermo:
we keep this sort of stuff in the [`tools/` dir](https://github.com/serpent-os/recipes/tree/main/tools) -- I think Joey already added a few helpers for bash?

Feel free to update and merge as appropriate.

Thanks!

ermo:
> I only tested the bash script and the experience wasn't not starightforward. In Solus, I can just symlink the script to `~/.bashrc.d/` and it will just work. In Serpent, I have to copy the `/usr/share/defaults/etc/profile` to `/etc/bashrc` and added this line (mimiicking what Solus does in its equivalent file):
> 
> ```
> if [ -d ~/.bashrc.d ]; then
>     for rc in ~/.bashrc.d/*; do
> 	if [ -f "$rc" ]; then
> 	    . "$rc"
> 	fi
>     done
>     unset rc
> fi
> ```
> 
> I have not tested the zsh and fish script. Cheers!

The real solution is to do this:
1. Add a `~/.bash_profile` that `~/.bashrc`
2. Add a `~/.bashrc` that first sources `/usr/share/defaults/etc/profile` and then adds the .bashrc.d stanza you have above.

We need to make this the default in the (the equivalent to) /etc/skel for Serpent at some point soon.

malfisya:
> The better solution might be this:
> 
>     1. Add a `~/.bash_profile` that does a `source ~/.bashrc`
> 
>     2. Add a `~/.bashrc` that first sources `/usr/share/defaults/etc/profile` and then adds the .bashrc.d stanza you have above.

This works. Thank you.

I moved the scripts to be under `tools/` directory. I don't want to disrupt the current workflow, so the function to navigate between package dir is still `chpkg` instead of `gotopkg` like in Solus. I still have not tested the zsh and fish script yet, as I am not feeling well these few days. Theoretically, they should work as they did in Solus. Cheers!

malfisya:
> Feel free to update and merge as appropriate.

Well, I updated it but merging? Come one ermo this is Serpent 😆

ermo (changes requested):
Please rename the bash script from `.sh` to `.bash`.

Looks good otherwise.

ermo (approved):
This looks like a decent start.

Thanks!

PR #446, opened 2024-12-17
asusd: Add at v6.1.0-rc2

flukejones:
Add asusd laptop control daemon for ASUS laptops.


Pheoxy:
You need to fix the homepage in the stone.yaml

ReillyBrogan (changes requested):
I was going to leave specific comments here, but honestly I'm very confused by this build.

- Why is it specifying `toolchain: gnu`, does it not build with Clang?
- Is mold a build requirement or is it just because it shaves a bit of time off the build? (hint: if I'm asking this question it's because there should be a comment in the recipe explaining the choice)
- Why is it pulling in `cmake` when it appears to use `make`?

ReillyBrogan:
Oh okay I just realized that you copied and pasted what was in the zed recipe. Feel free to ping me when you want this to actually be reviewed then.

a/asusd/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+homepage    : https://asus-linux.org/
+upstreams   :
+    - git|https://gitlab.com/asus-linux/asusctl.git : d785e17f9518c414134ce00585f1d83c0698090c
+description : |
+    Control ASUS platform features such as LEDs, AniMe, fans and more.
+builddeps   :
+    - binary(cargo)
+    - binary(cc)
+    - binary(clang)
+    - binary(make)
```
ReillyBrogan, line 21: These four dependencies can most likely be removed. 

- `clang` is pulled in automatically as a dependency of cargo/rust. 
- `cc` shouldn't be required if the makefiles/scripts are respecting the value of the CC and CXX environmental variables (and if they're not then most likely they're not respecting other variables like CFLAGS/CXXFLAGS/LDFLAGS and should be fixed). 
- Boulder will automatically add builddeps when certain macros are used. `%cargo_fetch` automatically adds a dependency on cargo, `%make` and `%make_install` add make.

a/asusd/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - pkgconfig(libinput)
+    - pkgconfig(libudev)
+    - pkgconfig(libseat)
+    - pkgconfig(libzstd)
+    - pkgconfig(libpcre2-8)
+    - pkgconfig(expat)
+    - pkgconfig(gbm)
+    - pkgconfig(gtk+-3.0)
+    - pkgconfig(xkbcommon)
```
ReillyBrogan, line 30: I feel like many of these pkgconfig dependencies are unnecessary and can be removed. I took a look at the project source code and (correct me if I'm wrong since you seem heavily involved in the project) it looks like it's only or mostly only Rust code. For rust code you can generally infer system library dependencies by checking the Cargo.lock and looking for `-sys` depenencies and dependencies that have `cc` or `pkg-config` as dependencies of themselves. While I see a few such dependencies many of them don't seem to correspond to the dependencies in the list here, nor are they getting linked against (see the `soname(*)` dependencies in `manifest.jsonc`).

a/asusd/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: 376375
+  rss: ~
```
ReillyBrogan, line 3: Since this is a Gitlab-hosted project you can grab the RSS feed URL from the tags page: https://gitlab.com/asus-linux/asusctl/-/tags

a/asusd/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+toolchain   : gnu # strip is used by default due to bin size
```
ReillyBrogan, line 31: This is still confusing to me, and the comment seems like you might think `toolchain: gnu` means something it doesn't. By default we setup environmental variables and add builddeps such that Clang/LLVM is used as the toolchain. `toolchain: gnu` changes that so that GCC and binutils (`ld`) are added as builddeps and environmental variables are set so that it's used as the default toolchain. 

It sounds like you are trying to use `toolchain: gnu` so that `strip` gets pulled in (when LLVM is used it prefers `llvm-strip` instead). This is probably incorrect for two reasons:
- The makefiles/scripts should check the value of the `STRIP` environmental variable and use the value of that instead of just looking for `strip` (you can have `strip` as a fallback if `$STRIP` isn't set)
- We actually almost never want the package recipe to run `strip` itself. Boulder actually runs strip itself at the end of the build on all detected ELF binaries and separates out the debug information so that `-dbginfo` subpackages can be created. This allows a user to install said debug info themselves so they have good stacktraces for crashes (and in the future we'll have debuginfod so debug information can be pulled in automatically).

a/asusd/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+networking  : yes
+setup       : |
+    %cargo_fetch
+build       : |
+    RUST_BACKTRACE=1 %make
+
+# the Makefile installs the bins + resources
+# so lets use that instead of doing it manually
+install     : |
+    %make_install
```
ReillyBrogan, line 41: I see that this uses Make to wrap cargo. If possible we want [our default cargo options](https://github.com/serpent-os/tools/blob/main/boulder/data/macros/actions/cargo.yaml) to be passed to Cargo directly, could you see if something like this will work? Escaping quotations might be difficult, in that case it might be prudent to just use `%cargo_build` directly in the build phase and just save `%make_install` for the install.
```
build       : |
    RUST_BACKTRACE=1 %make ARGS="%(options_cargo_release)" 
# the Makefile installs the bins + resources
# so lets use that instead of doing it manually
install     : |
    %make_install TARGET="%(target_triple)"
```

ReillyBrogan (changes requested):

a/asusd/stone.yaml
flukejones: the crate `slint` requires most of these.

a/asusd/stone.yaml
flukejones: Ah, I'll update the Makefile then

a/asusd/stone.yaml
flukejones: I'm using makefile to install resources like this https://gitlab.com/asus-linux/asusctl/-/blob/main/Makefile?ref_type=heads#L58

Reason I did it this way was to avoid having to manually do it for all package types.

flukejones:
I think that was a good clean up. Build tested, install tested, run tested.

ReillyBrogan (changes requested):
Looks much better now! The only thing that _needs_ to be fixed is that some changes to the linux desktop manifest seem to have snuck in during a rebase/amend.

Once that's done I'll approve this, but if you want since you need to poke it anyway you could move the comment about installing data/bins into the install block itself. The other minor thing you could change is to change
```
license     :
    - MPL-2.0
```
To
```
license     : MPL-2.0
```

Neither of these changes require a rebuild.

The final optional thing you could do would be to pre-enable the services out of the box (if that's desirable to users).

If you want to do that then create two files `asusd.preset` and `asusd-user.preset` in the pkgdir with the following contents respectively:
```
enable asusd.service
```
and
```
enable asusd-user.service
```
And then install these to `%(installroot)%(libdir)/systemd/system-preset/50-asusd.preset` and `%(installroot)%(libdir)/systemd/user-preset/50-asusd-user.preset`. Then they'll be enabled by default by a trigger.

Ignore if the services are DBUS-activated instead.

flukejones:
The services are udev rule activated. But it would be handy to be able to start them on install

ReillyBrogan (approved):

ikeycode (approved):

PR #452, opened 2024-12-20
bash: Enable support for sourcing from ~/.bashrc.d OOTB

joebonrichie:
**Test Plan**

- Ensure that the user shell in `/etc/passwd` is `/usr/bin/bash`
- Remove any pre-existing `~/.bash{rc,_profile}`
- `mkdir -pv ~/.bashrc.d`
- `echo 'export KILLROY_WAS_HERE=bash' > ~/.bashrc.d/99_KILLROY_WAS_HERE.bash`
- Open new terminal
- Check that  `echo $KILLROY_WAS_HERE` outputs `bash`


b/bash/pkg/profile
```diff
@@ -16,4 +16,13 @@ if [ -d /etc/profile.d ] ; then
     unset script
 fi
 
+if [ -d ~/.bashrc.d ]; then
```
ermo, line 19: This should probably add a "are we running bash"? conditional with `and` logic, to avoid the solus situation of zsh users getting tripped up over it?

Can add a similar .zshrc.d section too for good measure in a separate PR?

ermo (changes requested):
Need to consider the $OTHERSHELL use case here as well, hence requesting changes.

joebonrichie:
lgtm

ikeycode (approved):

PR #462, opened 2024-12-24
libiio: Add at v0.26

philmmanjaro:
Add libiio, a cross platform library for interfacing with local and remote Linux IIO devices .


l/libiio/stone.yaml
```diff
@@ -0,0 +1,28 @@
...
+summary: A cross platform library for interfacing with local and remote Linux IIO devices
+description: |
+  libiio is used to interface to the Linux Industrial Input/Output (IIO) Subsystem.
+builddeps:
+  - binary(bison)
+  - binary(flex)
+  - libaio-devel
+  - pkgconfig(avahi-glib)
+  - pkgconfig(libusb-1.0)
+  - pkgconfig(libxml-2.0)
```
ReillyBrogan, line 21: `pkgconfig` dependencies are ordered at the top of the list

l/libiio/stone.yaml
```diff
@@ -0,0 +1,28 @@
...
+license: LGPL-2.1 GPL-2.0
```
ReillyBrogan, line 22: `license` needs to be moved up above `summary`, and needs to be a valid SPDX license identifier. There are several examples in the repo.

ReillyBrogan (changes requested):

l/libiio/stone.yaml
philmmanjaro: Are you sure? when I looked as systemd for example you had the order binary, regular package names and last pkgconfig. kbd is similar ordered.

l/libiio/stone.yaml
philmmanjaro: It is not constant where the license line is placed:
ent: below description
inxi: below description
kdb: below description
itstool: below description
python-pyyaml: below description
zlib: below summary

ikeycode:
In theory the order should be whatever comes out of "boulder new" but truly without linting and matching docs it's not something we should be nitpicking?

Alpha sort dep lists with standard name deps last, has been our usual trend. The only thing that should be an issue for merge is ensuring license correctness, and again lints are still needed.

ikeycode (approved):

PR #463, opened 2024-12-24
inputplumber: Add at v0.40.0

philmmanjaro:
Add inputplumber, an open source input router and remapper daemon for Linux.

Merge after #462


i/inputplumber/stone.yaml
```diff
@@ -0,0 +1,34 @@
...
+description: |
+  InputPlumber is an open source input routing and control daemon for Linux.
+  It can be used to combine any number of input devices (like gamepads, mice, and keyboards)
+  and translate their input to a variety of virtual device formats.
+networking: yes
+builddeps:
+  - binary(cargo)
+  - binary(clang)
+  - binary(llvm-config)
+  - clang-devel
```
ReillyBrogan, line 22: Are these dependencies actually used? `binary(clang)` is unnecessary as it's the default toolchain on Serpent and is pulled in automatically, and I don't see that this is linking against libclang or libllvm which I'd expect for a clang-devel/llvm-config dependency.

ReillyBrogan (approved):
When creating a PR for a package that has dependencies on other packages please put all of your changes in the _same_ branch/PR. It looks like you opened a separate PR for libiio, since that library is newly added for inputplumber specifically it should be in this PR as well. Or to put it another way, would there be a point at all to merging libiio if we never merged inputplumber?

So close the libiio PR, cherry-pick the commit into this branch, re-order your commits so that they are in order of how they should be built by the build server, and then it will be structured correctly.

ReillyBrogan (dismissed):
Woops hit the wrong button

ReillyBrogan:
For example, take a look at my recent PR for mold: https://github.com/serpent-os/recipes/pull/447

Note how I add all of the new dependencies I needed in the same PR, each package change is a separate commit, and how the commits are ordered by how they need to be built by the build server. That's how this one needs to be as well.

i/inputplumber/stone.yaml
philmmanjaro: Without `binary(llvm-config)` it doesn't find llvm-config and without `clang-devel` the libclang.so is not found. I dropped binary clang.

ikeycode (approved):

PR #468, opened 2024-12-25
glibc/binutils: Update

ReillyBrogan:
Changes:
- Update glibc to v2.40 and update various build configs
- Update binutils to v2.43.0 and update various build configs
- Rebuild gcc and fix a rpath issue with newer boulder

Note that this also includes the blake3 package from https://github.com/serpent-os/recipes/pull/447 as that's needed by binutils.


ikeycode (approved):
lgtm =)

PR #478, opened 2024-12-27
Qt5: Switch back to KDE source, various fixes, add qtx11extras/qtwayland

ReillyBrogan:
Changes:
- Switch back to KDE source
- Add qt5-x11extras and qt5-wayland
- Various cleanups for devel
- Add a bunch of patches to improve Wayland integration
- Rundeps fixes


ReillyBrogan:
One sec, some of the version numbers are fucked

ReillyBrogan:
Fixed

q/qt5-base/stone.yaml
```diff
@@ -4,11 +4,11 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 name        : qt5-base
-version     : 5.15.16
-release     : 2
+version     : 5.15.16+kde.2529f7f0
```
ikeycode, line 7: this should be `+git.$shortRef` - `kde` isn't a VCS its just the upstream we're pulling from. If the concern is confusion around the *origin* - perhaps we also need a way to incorporate this?

ikeycode (approved):

PR #483, opened 2024-12-29
botan: Add at 3.6.1

Vixea:
Doesn't build with clang
Looks like everything is bundled in libbotan so no runtime deps needed

How would I add monitoring to it?


ReillyBrogan:
Marked as a draft until whatever this is a dependency of is added to the PR

Vixea:
I'll close this and add it when I'm done building keepassxc

PR #482, opened 2024-12-29
bash: Use a fan-in stateless configuration file

ermo:
This change should achieve the following:
- Enable all POSIX sh compatible shells to source the same stateless config file in their default startup scripts.
  - For bash login shells, this is /usr/share/defaults/etc/profile
  - For bash non-login shells, this is /usr/share/defaults/etc/bashrc
- Enable easy testing of drop-in profile.d/ or ~/.bashrc.d/ changes
  - It is essentially as simple as opening a new virtual terminal, which is very useful for packagers.
- Enable special casing for each supported shell
  - This ensures that we can set up each supported shell to be configured as conveniently as possible via relevant *.d/ directory drop-ins.
- Enable us to encourage users to keep bash as their login shell, given they can instead set up a ~/.bashrc.d/99-exec-my-interactive-shell.bash file.
  - This will exec users' preferred interactive shell using the already initialised environment, meaning we can get away with only maintaining *.sh format drop-in files in /usr/share/defaults/profile.d/
  
**Test Plan**

Login shell:

```
echo exit | strace bash -li |& grep '^open'
(...)
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/dev/tty", O_RDWR|O_NONBLOCK) = 3
openat(AT_FDCWD, "/usr/share/terminfo/x/xterm-256color", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/serpent-stateless-shell-conf.sh", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile.d/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile.d/10_default_PAGER_less.sh", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile.d/10-vte.sh", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile.d/40-prompt.sh", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/share/defaults/etc/profile.d/50-starship.sh", O_RDONLY) = 3
[...]

PR #492, opened 2024-12-29
fish: bash as login shell, fish as interactive shell

ermo:
This is a companion update to the #482 and #508 PRs (to be landed after both) , which piggybacks onto the configuration changes and offers a single file that can optionally start fish as an interactive shell after sourcing the default stateless shell settings via bash.

This file can either be symlinked into into ~/.bashrc.d from /usr/share/defaults/profile.d or, alternatively, be sourced from the user's custom ~/.bashrc file.

**Context for the proposed solution**
- https://wiki.gentoo.org/wiki/Fish#Caveats
- https://wiki.archlinux.org/title/Fish#Modify_.bashrc_to_drop_into_fish 
- https://nixos.wiki/wiki/Fish

**Test Plan**

- Install fish
- ln -sv /usr/share/defaults/profile.d/99-exec-fish-as-interactive-shell.conf ~/.bashrc.d/99-exec-fish-as-interactive-shell.conf.sh
- Open a new virtual terminal and see fish getting started as the interactive shell w/starship prompt enabled ootb if it is installed
- Execute `bash` and see bash get started with the correct environment

What this looks like (I am using a slightly customised fish prompt):

```
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
󰌽 ermo@snekws:~/.bashrc.d                                                                                                16:06
❯ 
```


ReillyBrogan:
This smells to me like using a sledgehammer to hammer a square peg into a round hole. It's standard to set your login shell with `chsh -s` which will completely bypass this, and if this is for things like setting environmental variables why aren't we using pam_env or system-environment-generators? And if it's for things like pretty-printing `ls`, well, fish users want fish to handle that.

ermo:
> It's standard to set your login shell with `chsh -s` which will completely bypass this

At that stage, the user is responsible for their startup scripts anyway, since they're changing defaults. 

The present solution literally enables you test fish out and still get the serpent default env, just by doing a single symlink command and no faffing with configurations (and no extra burden on package maintainers to maintain separate fish-format scripts w/defaults).

> and if this is for things like setting environmental variables why aren't we using pam_env or system-environment-generators?

For context, the parent PR of this solution works for both login and non-login shells (and ssh sessions), meaning that you can test the env var set by a locally installed package just by opening a new shell in a tab.

From what I can tell, `pam_env` is for login shells only?

As for `systemd-environment-d-generators`, the documentation makes it appear that env vars only reflect if something is started by a systemd service if I am reading correctly? So a new GNOME terminal (which starts a .service according to the docs) would likely reflect env var changes, but would a new tab...?

I could be misreading/misunderstading ofc, in which case, feel free to elaborate on the scheme you had in mind, and its UX for users (and packagers when testing new stuff).

Note that part of point of the parent PR is to avoid having to maintain skeleton files and their tendency of having to do 3-way-merges w/${HOME} dotfiles, once drift starts happening over time in that scenario.

> And if it's for things like pretty-printing `ls`, well, fish users want fish to handle that.

Not sure how you got the idea that this is for pretty printing `ls` in fish...?

PR #493, opened 2024-12-30
Rust: 1.83.0

ReillyBrogan:
Changes:
- Add libgit2 and libssh2 for Rust
- Add jemalloc for Rust
- Update nghttp2 and build 32bit for curl
- Build with curl with nghttp2 and disable hyper support (incompatible with http2 and now-unsupported upstream)
- Update Rust to v1.83.0
- Use jemalloc allocator for rustc
- Use more system libs with Rust/cargo (including libssh2 and libgit2)
- Update cargo-c to 0.10.7 and use system libsqlite, libgit2, and libssh2
- Build zed to use system libgit2, libsqlite, and libzstd.


ikeycode (approved):
will accept the zed change for the meantime but if it proves awkward in the long run we'll need to do something simpler

ikeycode (approved):

PR #495, opened 2024-12-31
golang: Fix goroot auto-detection

ReillyBrogan:
Having GOROOT set during the build causes said value to be emitted into the binary symbol table, which can apparently cause the automatic detection of GOROOT by resolving the /usr/bin/go symlink to fail. Ensure that it is unset during the build and now that it's no longer necessary we can remove the profile.d script setting it.


ermo (approved):
Tested successfully. Excellent job and thank you for chasing this down!

PR #505, opened 2025-01-01
linux-handheld: Add at 6.12.8

philmmanjaro:
This is an alternative kernel for Handheld devices. Currently tested on OrangePi Neo and ASUS ROG Ally.
At some point parts of the config and patches can be added to the regular `linux-desktop` package, if wanted.


ikeycode (approved):
🏆

PR #507, opened 2025-01-03
cosmic-session: Import from systemd user manager

ReillyBrogan:
As reported by users on Matrix Cosmic session has issues that are caused because environmental variables set by the systemd user manager are not being imported into the environment.


ikeycode (approved):

ikeycode:
Can we try and upstream the patch to the pop-os guys please?
See: https://github.com/pop-os/cosmic-session/issues/76

If time is constrained I'm happy to rebase and push for you

PR #514, opened 2025-01-06
opengamepadui-installer: Add at 20240730

philmmanjaro:
adds a nice graphical installer for handheld devices based on Godot game engine


ikeycode (changes requested):
none of the builddeps are actually used.. ? And it's using a binary prebuilt elsewhere, not using our toolchain or linking defaults..

philmmanjaro:
@ikeycode we might want to add rundeps similar to OGUI here as well ...

philmmanjaro:
Should be ready now @ikeycode

ikeycode (approved):

PR #515, opened 2025-01-06
Enable support for reading .stone files in file-roller

joebonrichie:


ReillyBrogan:
Note that as mentioned in Solus the libstone patch for libarchive currently breaks the test suite (I think around zip files IIRC). This has to do with moving the file position I think.

ikeycode (approved):

PR #520, opened 2025-01-06
Package gtk docsy things

joebonrichie:
Tested with msgraph and upower

**Packager's Notes**
- `gi-docgen` found an edge case with our python providers we need may need to resolve


ikeycode:
after merge ill sync python provider normalisation

ikeycode (approved):

PR #525, opened 2025-01-07
linux-handheld: Adjust package

philmmanjaro:
- re-arrange patches
- re-use clang


ikeycode (approved):
<3 back to llvm comfort land

PR #527, opened 2025-01-08
linux-desktop: update asus patch series

flukejones:
Refreshes the asus patch series with a few fixes and corrections for stability under clang.


ikeycode (approved):
thanks - ill get this bumped later today =)

PR #533, opened 2025-01-10
starship: Switch to bash/zsh default PS1 style

ermo:
People seem to like my bash PS1, so figured I might as well put up a PR for making starship look similar.

image

Note that the proposed prompt also looks cleaner in `text blocks`:

```
󰌽 ermo@blueshift:~/repos/serpent-os/recipes/s/starship [  starship-prompt-matching-default-bash-and-zsh-PS1 ? ]          20:02
❯
```

(I'm sanguine about this being closed if the changes aren't acceptable)


ermo:
Doesn't look like there is much love for this, so closing.

PR #535, opened 2025-01-10
llvm: Cleanups and improvements

ReillyBrogan:
Changes:
- Shrink the 32bit build significantly by limiting the target support to X86 and AMDGPU, and also only building LLVM. Also disable several additional features not needed for the 32bit build
- And ARM and WebAssembly target support to the 64bit build for users cross-compiling
- Get ccache working properly, and add debugging information so that it's possible to see how effective it is.
- Disable building with PGO (this is temporary and will be re-enabled post LLVM 19)
- Add openmp support (libomp)
- Add clang-tools-extra (clang-tidy plus many other tools) so that we can build a more generally useful LLVM distribution for users.
- Create `clang-tools` subpackage and move all of the build artifacts of `clang-tools-extra` there as well as clang-analyzer and other tools that were in `clang` that were not strictly necessary for compilation
- Ensure that subpackage rundeps effectively pull in the packages they need. For instance `clang-devel` now depends on `llvm-devel`
- Update various package descriptions
- Build lld, clang, and llvm-ar so that they use the mimalloc memory allocator (improves performance by ~7%)
- Patch LLD so that it always behaves as though `--build-id=sha1` was passed (this matches binutils/mold)
- Patch LLD so that it always behaves as though `--compress-debug-symbols=zstd` was passed (this matches binutils/mold)

Depends on serpent-os/tools#402
Depends on serpent-os/recipes#447

Depends on serpent-os/tools#381 (not technically required but I only tested this with that PR applied)


ikeycode (approved):
lets go

PR #540, opened 2025-01-11
linux-desktop: Update asus patch series

flukejones:
Use a separated patch series the same as in linux-handheld package. This will make it easier to see if there are any significant differences between desktop and handheld in relation to ASUS handhelds specificaly.


flukejones:
Also makes it easier to review patch changes in future I guess.

ikeycode (approved):
thank you!

PR #547, opened 2025-01-13
linux-desktop: add common USB-to-serial drivers

flukejones:
Compile a few USB-to-UART (serial) drivers which are commonly used for embedded environment debugging.


flukejones:
I require these drivers for work. It would be good to have them in the base build of serpent so I can also continue to use the pre-built nvidia drivers.

philmmanjaro:
LGTM

ikeycode (approved):

ikeycode:
ack no harm in it. i'll include in next kernel update =)

PR #554, opened 2025-01-17
LLVM v19.1 Part 1

ReillyBrogan:
### DO NOT MERGE
Do not merge this PR until Part 2 (the rebuilds) is uploaded and approved. This PR can and should be approved but not merged.

Changes:
- Updates libcxx, llvm, and llvm-bolt to v19.1.6
- Bootstraps ldc and rust against LLVM 19.1

Note that in LLVM the `libomptarget` build artifacts were split out of openmp and into a new llvm project `offload`. Unfortunately it's only buildable with `LLVM_ENABLE_RUNTIMES`, but due to how our toolchain is setup (with libcxx being built first) it's very difficult to get that working without extensive patching. For that reason I've opted not to build it for now, if/when it's required we can add it as a standalone build like lldb, libclc, or llvm-bolt.


ikeycode:
👻 👍️

ikeycode (approved):

PR #568, opened 2025-01-24
mesa: Update to v24.3.4

philmmanjaro:
- also add render fix for Chromium based browsers


ikeycode (approved):
oh - love the preemptive fix =)

PR #574, opened 2025-01-25
freetype: enable brotli

Vixea:


Vixea (commented):
should probably mention this but I need this for blender(probably not going to push blender upstream yet) but since freetype is already packaged this should be harmless

joebonrichie:
Only reason i haven't merged this is there is a nasty circular dependency issue between freetype/harfbuzz that summit can potentially shit the bed on which may need manual admin intervention

Vixea:
Ahh fun

Vixea:
I should probably rebuild every package with freetype anyway, would that help at all?

PR #577, opened 2025-01-30
Add SDL3 and switch to sdl2-compat

ReillyBrogan:
- Add SDL3 (relatively complete build, several libraries are strategically chosen to be linked against rather than dlopen()ed)
- Deprecate SDL2
- Add sdl2-compat
- Rebuild everything against it


s/sdl2/stone.yaml
```diff
@@ -14,23 +14,16 @@ description : |
...
-    - pkgconfig(xext)
-    - pkgconfig(xkbcommon)
-setup       : |
-    %cmake
-build       : |
-    %cmake_build
 install     : |
-    %cmake_install
+    %install_dir %(installroot)/%(datadir)/tmp
+    touch %(installroot)/%(datadir)/tmp/%(name){,-devel,-dbginfo}
```
ikeycode, line 19: this doesnt seem consistent with other renames like uutils-coreutils.. ?

ReillyBrogan: Should be better now? I'd note that the previous way is how I did it in GCC

ikeycode (approved):
ty ty sexy stuff here

PR #579, opened 2025-01-30
Added wf-recorder

Ghodawalaaman:


ermo (changes requested):
Nice first effort.

Just a small note that (per https://github.com/serpent-os/recipes?tab=readme-ov-file#git-summary-requirements) it would be nice if you could please change the commit title and the PR title so it reads like this:

> wf-recorder: Add at v0.5.0+git.560bb92

In addition, it'd be a nice touch if you included a link to the video you made with wf-recorder in a \*\*Test Plan\*\* section in the PR (makes it super obvious that you tested it and that it works ootb)

>
> **Test Plan**
> - Build wf-recorder recipe
> - Install wf-recorder from my local repo 
> - Successfully use wf-recorder to create a video (link)
>

PR #587, opened 2025-02-04
[NFC] gdm: add Serpent OS logo

lz-coder:
I think it looks cool :)


ikeycode:
This should already be handled in `serpent-gnome-defaults` package, was it not working for you?

lz-coder:
> This should already be handled in `serpent-gnome-defaults` package, was it not working for you?

Ooh, its not installed as default for me...

Hmm ... I uninstalled some packages provided by **serpent-gnome-defaults**: gnome-browser-connector, appindicator and tilingshell extensions, and papirus-icon-theme  ( i like vanilla Gnome... ).

So if i want the logo on GDM, i need to install serpent-gnome-defaults, even if i don't want that other packages on my system, Its not a problem for me if its the "Serpent OS' way of doing things :) 
I'm ok with that...

ikeycode (approved):

PR #592, opened 2025-02-05
font-noto-cjk: fix file permissions

philmmanjaro:


philmmanjaro:
my mistake. tested an outdate install image.

PR #598, opened 2025-02-09
amdgpu_top: Adding at v0.10.3

cybin:


a/amdgpu_top/stone.yaml
```diff
@@ -0,0 +1,21 @@
+name        : amdgpu_top
```
ikeycode, line 1: is this hand generated? missing the spdx header

ikeycode (approved):

PR #608, opened 2025-02-10
Libunwind

ryzendew:
just trying to help

This is used by 
- wine
- GDB
- LLVM
- valgrind
- OpenSSH


philmmanjaro:
Seems not needed at all since we use libunwind provided by llvm

ryzendew:
> Seems not needed at all since we use libunwind provided by llvm

i had llvm installed and it was not seen by wine at all

ikeycode:
```
❯ moss info 'soname(libunwind.so.1(x86_64))'
Name                 libcxx
Status               Installed
Version              19.1.7
Release number       12
Homepage             https://www.llvm.org
Summary              Modern C++ standard library
Description          A modern C++ standard library from the LLVM project, aiming
                     for minimal overhead and compilation times. Targets the
                     C++11 standard. This is the default C++ library in Serpent
                     OS used in conjunction with the LLVM toolchain.

Dependencies         • soname(ld-linux-x86-64.so.2(x86_64))
                     • soname(libc.so.6(x86_64))

Providers            • name(libcxx)
                     • soname(libc++.so.1(x86_64))
                     • soname(libunwind.so.1(x86_64))
```

`libcxx-devel` provides libunwind-devel

PR #610, opened 2025-02-10
Adds p7zip also known as 7z

ryzendew:
Winetricks needs this for some files


p/p7zip/justfile
```diff
@@ -0,0 +1,132 @@
+#
+# SPDX-FileCopyrightText: © 2020-2025 Serpent OS Developers
```
Vixea, line 2: This isn't used anywhere and shouldn't be needed this should be deleted.

p/p7zip/stone.yaml
```diff
@@ -0,0 +1,24 @@
+name: 7zip
+version: 24.09
+release: 1
+summary: 7-Zip command-line version for Linux
+license: LGPL-2.1-or-later
```
Vixea, line 5: It's also licensed under bsd-3. maybe that part isn't built by default?

ryzendew: added it anyway

p/p7zip/stone.yaml
```diff
@@ -0,0 +1,24 @@
+name: 7zip
+version: 24.09
+release: 1
+summary: 7-Zip command-line version for Linux
+license: LGPL-2.1-or-later and BSD 3-clause License
```
Vixea, line 5: That's not how you add a multiple licensed package,

BSD-3-Clause 
On a new line and without and, and license

Vixea: As it must be formatted like the identifiers in this list
https://spdx.org/licenses/

p/p7zip/stone.yaml
```diff
@@ -0,0 +1,25 @@
+name: 7zip
+version: 24.09
+release: 1
+summary: 7-Zip command-line version for Linux
+license: LGPL-2.1-or-later
+license: BSD-3-Clause
```
Vixea, line 6: There should only be one license: 
Delete it and add enough spaces to align to the start of the lgpl one

ryzendew:
i need to figure out how to compile Format7z

PR #618, opened 2025-02-12
syncthing: Add at 1.29.2

lz-coder:


ermo:
@ikeycode What do we want to do with LICENSE files? Currently, recipes aren't really well aligned:

1. install them to `%(datadir)/licenses/%(name)/<the thing>` (see [f/fdk-aac/stone.yaml](https://github.com/serpent-os/recipes/blob/main/f/fdk-aac/stone.yaml#L23))
1. install them to `%(docdir)/%(name)/<the thing>` (see [n/ntpd-rs/strone.yaml](https://github.com/serpent-os/recipes/blob/main/n/ntpd-rs/stone.yaml#L39-L40))

The present recipe uses the first approach.

ikeycode:
id say docdir

ikeycode (approved):

Issue #634, opened 2025-02-18
Samba package is broken


philmmanjaro:
When I look into `systemctl` I find the following issues based on the **samba** package:

```
× samba.service - Samba AD Daemon
     Loaded: loaded (/usr/lib/systemd/system/samba.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2025-02-18 00:15:16 PST; 1h 35min ago
       Docs: man:samba(8)
             man:samba(7)
             man:smb.conf(5)
    Process: 1523 ExecStart=/usr/sbin/samba --foreground --no-process-group $SAMBAOPTIONS (code=exited, status=203/EXEC)
   Main PID: 1523 (code=exited, status=203/EXEC)
        CPU: 467us

Feb 18 00:15:16 snekbox systemd[1]: Starting Samba AD Daemon...
Feb 18 00:15:16 snekbox (samba)[1523]: samba.service: Failed to locate executable /usr/sbin/samba: No such file or directory
Feb 18 00:15:16 snekbox (samba)[1523]: samba.service: Failed at step EXEC spawning /usr/sbin/samba: No such file or directory
Feb 18 00:15:16 snekbox systemd[1]: samba.service: Main process exited, code=exited, status=203/EXEC
Feb 18 00:15:16 snekbox systemd[1]: samba.service: Failed with result 'exit-code'.
Feb 18 00:15:16 snekbox systemd[1]: Failed to start Samba AD Daemon.
```

```
smb.service - Samba SMB Daemon
     Loaded: loaded (/usr/lib/systemd/system/smb.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2025-02-18 00:15:16 PST; 1h 35min ago
       Docs: man:smbd(8)
             man:samba(7)
             man:smb.conf(5)
    Process: 1542 ExecStart=/usr/sbin/smbd --foreground --no-process-group $SMBDOPTIONS (code=exited, status=1/FAILURE)
   Main PID: 1542 (code=exited, status=1/FAILURE)
        CPU: 22ms

Feb 18 00:15:16 snekbox systemd[1]: Starting Samba SMB Daemon...
Feb 18 00:15:16 snekbox (smbd)[1542]: smb.service: Referenced but unset environment variable evaluates to an empty string: SMBDOPTIONS
Feb 18 00:15:16 snekbox systemd[1]: smb.service: Main process exited, code=exited, status=1/FAILURE
[...]

PR #640, opened 2025-02-20
samba: don't install service files

philmmanjaro:
This would improve https://github.com/serpent-os/recipes/issues/634. The only service which would work would be for Samba NMB Daemon. However, mostly its service is disabled on most distributions.

From the Arch wiki: `If you want to make your server accessible via NetBIOS host name, set the desired name in the netbios name option in smb.conf and enable/start nmb.service.`


ikeycode (approved):
yeah the error logs aren't a good look.

PR #642, opened 2025-02-20
wine: Add at 10.2

cybin:


w/wine/stone.yaml
```diff
@@ -0,0 +1,77 @@
+#
+# SPDX-FileCopyrightText: © 2020-2025 Serpent OS Developers
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : wine
+version     : 10.1
+release     : 1
+homepage    : https://dl.winehq.org/wine/source/10.x
```
Vixea, line 9: maybe change this to just be https://www.winehq.org/

cybin: Oh yes, done.

ReillyBrogan:
I'm cleaning this up right now, don't merge.

ikeycode:
<3

cybin:
I think we should stick to the default behaviour, in regards to x11 and wayland. According to the release notes for 10.0, wine prefers x11 and you can switch to wayland by unsetting DISPLAY. Now that wayland is preferred, I wouldn't know what to do to force it back to x11. The wayland driver is known to cause problems, at least with CEF based UIs. I couldn't run the RSI launcher, for example. It's simply a white or black square with the wayland driver.

PR #651, opened 2025-02-22
asus kernel patch updates

flukejones:


philmmanjaro:
We have to think when it is best to update to 6.13 kernel series. There are still some issues reported on stable kernel mailing list

ikeycode (approved):
to ensure we're not overworking luke we'll shift branch

PR #682, opened 2025-03-07
Pipewire v1.4.1

philmmanjaro:
- libebur128: Add at v1.2.6
- pipewire: Update to v1.4.1

[Changelog](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/df1c36aec2aaf9e2eb1596b6b55e72c15fb2088a/NEWS) of pipewire 1.4, [Changelog](https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/1.4.1) of pipewire 1.4.1


ReillyBrogan:
libeburl128 is also in AerynOS/recipes#670, could you drop that commit and just mark that this PR depends on 670? The other one includes a monitoring.yaml. Also, it looks like the cmake arguments you have are all unnecessary since they aren't in the other attempt and the manifest is identical.

philmmanjaro:
seems there are some reports of regressions in 1.4 compared to 1.2. lets see.

ReillyBrogan (approved):

PR #697, opened 2025-03-18
sassc: Add at v3.6.2

joebonrichie:
Initially included for libadwaita 1.7.x


l/libsass/stone.yaml
```diff
@@ -0,0 +1,23 @@
...
+description : |
+    A C/C++ implementation of a Sass compiler
+license     : MIT
+setup       : |
+    %reconfigure --disable-static
+build       : |
+    %make
+install     : |
+    %make_install
+# hardcoded to gcc-ar
```
ReillyBrogan, line 22: I looked into this and the issue was actually in slibtool. I added a patch to it to fix it, can you remove this line and add a depends on https://github.com/AerynOS/recipes/pull/709 instead? You'll need to rebuild this package to pick up that it's linking to libc++ and libunwind (I tested it and it builds with updated slibtool).

ReillyBrogan (changes requested):

l/libsass/stone.yaml
ReillyBrogan: Also if binutils ar is the only thing you need from the gnu stack it's better to use `binary(ar)` as a builddep.

ReillyBrogan (approved):

PR #705, opened 2025-03-19
oniguruma: fix name

AMDmi3:


ikeycode:
Hey, out and about atm but this'll need some transition path, I'll help do that later:)

AMDmi3:
This is blocking repology parsing, I'll have to remove aerynos from repology if it's not fixed soon.

AMDmi3:
The policy is not to ignore individual packages, as that hides the problem and produces incomplete and unexpected results.

ikeycode:
> This is blocking repology parsing, I'll have to remove aerynos from repology if it's not fixed soon.

I was in a hospital with my wife for nigh a week. This wasn't exactly top of my priority list.

AMDmi3:
Please take your time, I hope you and your wife are well. I've just temporarily disabled AerynOS in Repology.

PR #708, opened 2025-03-19
GNOME 48

joebonrichie:
https://release.gnome.org/48/

Depends on #697
Depends on #698
Depends on #699
Depends on #700
Depends on #701
Depends on #702
Depends on #703


g/gnome-system-monitor/stone.yaml
```diff
@@ -18,6 +18,7 @@ builddeps   :
     - binary(lld)
     - binary(msgfmt)
     - binary(python3)
+    - pkgconfig(catch2-with-main)
```
ReillyBrogan, line 21: This would be a checkdep no?

joebonrichie: The catch2 headers are used during build, not just a checkdep.

ikeycode (approved):

PR #718, opened 2025-03-22
Sway/Wlroots: Update to latest

ReillyBrogan:
- Update wlroots to v0.18.2
- Update swaybg to v1.2.1
- Add wmenu at v0.1.9 (default sway menu)
- Update sway to v1.10.1

Note that this is untested but people who actually use Sway have not replied to my comment asking for testers so I suppose we'll have to scream-test this.


ikeycode (approved):
"scream test" 😆

PR #726, opened 2025-03-24
libbpf: Add at v1.5.0

ReillyBrogan:
Changes:
- Add libbpf
- Rename perf package to linux-tools and build other tools as subpackages
- Add bpftool
- Update iproute2 and support BPF
- Build systemd with clang and support BPF

This is largely scaffolding for BPF at this point, many BPF options need to be enabled in the kernel in order for any of this to be usable, but at least it won't hurt to have this in before that.

Depends on aerynos/recipes#716


ReillyBrogan:
Draft due to ABI changes and rebases, going to update this next week.

PR #729, opened 2025-03-25
llvm: Build runtimes and additional changes

ReillyBrogan:
Changes:
- Merge libc++/libunwind build into llvm package
- Merge llvm-bolt build into the llvm package
- Update Clang version string to AerynOS
- Support aerynos target triple
- Build compiler-rt as runtime instead of as project
- Build openmp as runtime instead of as project
- Enable offload runtime
- Move clang-scan-deps to clang package
- Add libcxx typeinfo fix for qtwayland
- Enable riscv target support
- Enable SPIR-V target support

Also, as mentioned in Matrix this needs to be carefully merged due to ABI incompat.

Depends on AerynOS/recipes#683


ReillyBrogan:
Known packages that need to be rebuilt after this:
- fmt
- qt6-core
- qt6-wayland (needs to be rebuilt due to libcxx typeinfo change)

ikeycode (approved):

PR #732, opened 2025-03-25
gdm: Split out libs

ReillyBrogan:
Split out the libs from GDM so that SDDM and other login managers can launch a gnome session without having to install GDM.


philmmanjaro (approved):
LGTM

ikeycode (approved):

PR #735, opened 2025-03-27
Build more packages with clang

ReillyBrogan:
- Add lzip so tar can decompress .lz archives
- Update ed to 1.21.1
- Get accountsservice, avahi, colord-gtk, cups, cyrus-sasl, ed, and elfutils all building with clang


ikeycode (approved):
love it

PR #737, opened 2025-03-28
Add just to build-essential package

davidsalter:
The "just" package is not present within the "build-essential" package, however the documentation states that  "just" should be installed as part of "build-essential" (https://aerynos.dev/packaging/workflow/prerequisites/)

This change adds "just" to the "build-essential" package.


ikeycode (approved):
good catch!

PR #757, opened 2025-03-29
Python updates

joebonrichie:


p/python-hatchling/manifest.x86_64.jsonc
```diff
@@ -4,30 +4,25 @@
...
-				"python-trove-classifiers",
-				"python-wheel"
+				"python(editables)",
+				"python(pathspec)",
+				"python(pluggy)",
+				"python(trove-classifiers)"
 			],
 			"depends": [
 				"binary(python3)",
-				"python(editables(3.11))",
```
ReillyBrogan, line 18: Does it still need editables as a builddep if it's no longer a rundep?

joebonrichie: nice catch

ReillyBrogan (approved):

PR #760, opened 2025-03-29
LLVM 20

ReillyBrogan:
Part of AerynOS/recipes#759

Depends on AerynOS/recipes#826
Depends on AerynOS/recipes#780


ermo (approved):
"Once more unto the breach, dear friends, once more"

PR #765, opened 2025-04-01
llvm: Fix cmake exports

ReillyBrogan:
Building against llvm itself was broken because the cmake files contained references to static libs that we had deleted. Fix this by creating a distribution in LLVM with a hand-picked list of components (that doesn't include unnecessary static libs) which ensures that the cmake files don't reference things that were deleted (because we just don't install them instead).

This has the beneficial side-effect of reducing the build times by a fair chunk since `ninja distribution` only builds things that are going to be installed (and their dependencies) which saves time building a few things we didn't include anyway.


ikeycode (approved):

ikeycode:
magic

PR #768, opened 2025-04-01
Qt6: v6.9.0

ReillyBrogan:
- Updated Qt6 stack to v6.9.0

Depends on AerynOS/recipes#775


ReillyBrogan:
Draft because 6.9.0 is supposed to be out tomorrow and I might as well do that one.

philmmanjaro:
Plasma 6.4 is still recommending 6.8 series of Qt. What are the benefits of going to 6.9?

philmmanjaro:
Seems 6.9.0 is scheduled for this week and 6.9.1 for May: https://wiki.qt.io/Qt_6.9_Release. Initial release may have unknown bugs ...

ReillyBrogan:
> Plasma 6.4 is still recommending 6.8 series of Qt. What are the benefits of going to 6.9?

I can drop the entire font patch series and it means I don't have to go through the major update with other Qt components when they're added. Also, that's just the minimum version Plasma supports it'll work with newer versions (though it may need fixes).

> Seems 6.9.0 is scheduled for this week and 6.9.1 for May: [wiki.qt.io/Qt_6.9_Release](https://wiki.qt.io/Qt_6.9_Release). Initial release may have unknown bugs ...

Yes, new Qt major versions always have bugs. IMO it's fine since by the time anyone is really using Plasma most of those will be sorted upstream.

ReillyBrogan:
Draft due to ABI changes and rebases, going to update this next week.

PR #777, opened 2025-04-03
gcc: Update to Aeryn triple

ReillyBrogan:
Depends on AerynOS/os-tools#450


ermo (approved):
We'll push this to main and then rebase the infra-alpha2-fixes branch on top of it for Rust infra testing and see what gives.

Thanks!

PR #780, opened 2025-04-03
llvm: Rebuild with updated Aeryn triple

ReillyBrogan:
Depends on AerynOS/os-tools#450
Depends on AerynOS/os-tools#451
Possibly depends on AerynOS/recipes#777


ermo (approved):
The prerequisites are in now, so let's get this merged.

Thanks!

PR #783, opened 2025-04-04
Jsoncpp: Add at 1.9.5

cybin:
This is part of waybar dependencies.


ermo (changes requested):
Please change the double quotes "some.version" to single quote 'some.version' (looks nicer).

Also, would you mind rebasing this onto the [2025-05-repo-rebuild](https://github.com/AerynOS/recipes/tree/2025-05-repo-rebuild) branch so we can get it in with the new infra?

cybin:
Will be added to the waybar PR.

PR #784, opened 2025-04-04
Iniparser: Add at 4.2.6

cybin:
This is part of waybar dependencies.


cybin:
Will be added to the waybar PR.

PR #788, opened 2025-04-05
inputplumber: Update to v0.52.1

philmmanjaro:


ReillyBrogan (approved):

philmmanjaro:
IP v0.52.0 has a new "deck-uhid" target device. It will really only work for the Deck, Ally/X, Go/S, and Zone, but it doesn't need the suspend service anymore. v0.52.1 adds more bugfixes to the mix.

ikeycode (approved):

PR #796, opened 2025-04-12
wireplumber: Update to 0.5.8

hphilm:
this also adds a patch for steam to avoid audio issues: https://github.com/ValveSoftware/steam-for-linux/issues/11901


ermo (approved):
LGTM, thanks!

PR #797, opened 2025-04-13
xz: Update to v5.8.1

joebonrichie:


hphilm (approved):
Needed for security reasons.

PR #802, opened 2025-04-15
nushell: update to 0.104.1

Yethal:


ReillyBrogan (changes requested):
You need to increase the release number and actually build this. Then test it on your system and commit the manifest changes.

Yethal:
@ermo PR stayed open for so long that nushell was updated in the meantime so I updated the PR as well, I hope it's not a problem.

ermo (changes requested):
Please run `just bump` and `just build` to rebuild it. 

After that is done, run `git add <the changed files>`, then commit the changes with `git commit --amend`, and finally run `git push -f` on your branch.

The release number should be `3` now, not `2` (it looks like the old version was bumped to release 2 on the rebuild branch already).

You're almost there. =)

PR #809, opened 2025-04-27
linux-pam: setuid root on unix_chkpwd

cybin:
Cosmic-greeter fails to unlock displays without it.


hphilm:
You need to bump `release     : 9` to `release     : 10` to have it rebuild ...

ermo:
@cybin : Please rebase and squash the commits, and make the title of the resulting single commit the same as the title of the PR.

l/linux-pam/stone.yaml
```diff
@@ -66,3 +66,6 @@ install     : |
 
     %tmpfiles "# Create run/faillock directory"
     %tmpfiles "d /run/faillock 0755 root root -"
+
+    chmod u+s %(installroot)/usr/sbin/unix_chkpwd
```
ermo, line 70: Please add a comment above so people don't need to git blame to understand why it is necessary to use setuid here.

ermo (changes requested):
A small nit, otherwise looks good.

ermo (approved):
LGTM, thanks!

PR #821, opened 2025-05-04
glibc: Update build deps

hphilm:
Fixes the issue that `zstd` is not found during installation routine ...


ermo (approved):
LGTM, thanks!

PR #824, opened 2025-05-08
libssh: Update to v0.11.1

hphilm:
Changing the toolchain to gnu fixes the error: `Your system must provide a __func__ macro`. So there is some issue when using clang ...


ermo (approved):
LGTM, thanks!

PR #826, opened 2025-05-08
ldc: Update to v1.40.1 and fix build

ReillyBrogan:
Additional changes:
- Adapt to LLVM target changes
- Update patterning to split out runtime libs and move all devel files back into main package

Depends on AerynOS/recipes#780


ermo (approved):
Fixes an annoying infra-alpha2-fixes branch build failure, so thanks!

PR #831, opened 2025-05-11
podman: Add at 5.4.2

### LINKED ISSUE #565: [Package Request]: Podman
### Existing package request?

### Homepage of package

https://podman.io/

Drop-in Replacement for Docker replacement which also provides rootless containers. Helpful on immutable/atomic distros

If i have missed something or docker is preferred then this can be resolved :)

### Licence of package?

Apache 2.0

https://github.com/containers/podman/releases/tag/v5.3.2

LDprg:
This PR aims to add podman and its dependencies:
 - podman: Add at 5.4.2
 - podman-compose: Add at 1.4.0
 - python-dotenv: Add at 1.1.0
 - yail: Add at 2.1.0
 - conmon: Add at 2.1.13
 - crun: Add at 1.21

It's not completely ready, since I am sure not how to handle default configs in aerynos? (Basic building works)

Also, it should be discussed, where #830 should use docker or podman (if there should be a dependency since it works with both).

Another thing is I want to include podman-docker as separate package (mainly for replacing docker commands). However, I didn't know how to create conflicting packages (podman-docker and docker can't be installed at the same time) and split the podman-docker and podman build correctly (since they are build from the same source)

This PR is a draft mainly to discuss how things should be done correctly.

Closes #565


ermo:
I really hate to say this, but we don't have conflicts support in moss yet. Additions to moss are blocked on versioned repos (infra alpha3 milestone), which is in turn blocked on the Rust infra port (infra alpha2 milestone, currently in progress and nearing completion).

Hence, it will be a while before we can land this. :pensive:

LDprg:
Oh, ok. I guess it is best to simply not package podman-docker (since it is not that important for now and is not required for podman itself). 

And for distrobox it is probably best to do it like arch linux does, simply set none of docker and podman as dependency. 
Or we just distribute two flavors like distrobox-docker, distrobox-podman and maybe a default one (I guess the podman one).

ermo:
> Or we just distribute two flavors like distrobox-docker, distrobox-podman and maybe a default one (I guess the podman one).

My gut says "ship both".

For the state the tooling is in currently, it probably does not make sense to set either as "the default".

ermo:
@LDprg You still around? Is this ready as is from your perspective?

LDprg:
@ermo I am still around (sadly currently very busy). This PR still not ready since default config for podman is still missing (maybe there is something else missing, but that's the last thing I remember that I was working on).

It will take some time (at least 3 weeks) until I have time to work on this again. So if anyone wants to continue this in the meantime feel free to do so.

PR #832, opened 2025-05-12
tools/heplers: fix error when not using symlinks

LDprg:
I noticed when using fish as shell that the helpers scripts won't allow directly loading commands with source.
The reason is that `(status -f)` does not provide a valid input for `readlink`, if symlinks are not used. Adding `-m` causes `readlink` to not validate the path (which is fine, because `dirname` does).

This is quite nice since I don't need to create a symlink using fish. (I can directly add `source path-to-script/tools/helpers.fish` in the config fish).


ermo (approved):
Nice explanation and feature/fix. LGTM, thanks!

PR #834, opened 2025-05-13
nvim: Add at 0.11.1

### LINKED ISSUE #627: [Package Request]: neovim
### Existing package request?

### Homepage of package

https://neovim.io/

You already offer vim, but nVim is an alternative written in Lua, it is nowadays preferred by the ricing community there are more diverse plugins systems. More performant. 

### Licence of package?

Apache-2.0

https://github.com/neovim/neovim/releases/download/v0.10.4/nvim-linux-x86_64.tar.gz

LDprg:
Closes #627
Related #353

This PR adds Neovim and its dependency. This is very much WIP, however it builds fine (currently not installable, see below). 

Somehow I can't get runtime deps to work correctly since moss automatically add `"soname(/usr/lib/lua/5.1/lpeg.so(x86_64))",` which can't be found since it should either be `"soname(lua/5.1/lpeg.so(x86_64))",` or `lua51-lpack,`.

Feel free to leave suggestions and feedback, especially about handling multiple Lua versions and package names (and maybe even a solution to the problem above). Just note that I will refactor this.


ReillyBrogan:
I'll try to get to this once all the toolchain stuff is out of the way, and the repo is fully rebuilt (see discussions in packaging channel for more info).

Regarding lua, does this work with luajit? That has lua 5.1 compatibility so maybe?

LDprg:
@ReillyBrogan Oh thx. I removed lua51 and used luajit instead (everything still builds fine). I wasn't aware of the compatibility.

However, the path problem with the lua library persists (exact the same error). I also tried a different machine for testing (just in case), same results.

ermo:
@ReillyBrogan Did you ever get around to this? Is there any issue that we need to link it to?

PR #838, opened 2025-05-15
strace: Update to v6.14 and fix build with clang

ReillyBrogan:


ermo (approved):
One more build failure resolved. Thank you!

PR #842, opened 2025-05-16
nano: Update to v8.4

sammypanda:
Hi! I don't use Nano but i was following through the packaging guide which included trying it out on the nano package and noticed 8.4 was available so thought maybe i could.

Go easy on me im nervous about these things and it seemed like the tests passed :sweat_smile:


ermo (changes requested):
Well done.

Would you mind rebasing this onto the [2025-05-repo-rebuild](https://github.com/AerynOS/recipes/tree/2025-05-repo-rebuild) branch so we can get it in with the new infra?

sammypanda:
wait i think i still rebased it wrong, it shouldn't say im trying to merge all those commits right? Why are there so many different ways to do it :sob: maybe i should just close this pr, remove the branch, recreate it and add the package on top of that.

ermo (approved):
LGTM, thanks!

PR #844, opened 2025-05-17
libcbor: Add at v0.12.0

d5ng4i:
Required for openssh FIDO support.

PR #846, opened 2025-05-17
libfido2: Add at v1.16.0

d5ng4i:
Required for openssh FIDO support.

We should also add the user created by lichen to `plugdev` [here](https://github.com/AerynOS/lichen/blob/990a7913eb69341233551f59ca13da437612d1ad/crates/installer/src/steps/postinstall.rs#L65). But `plugdev` exists only when libfido2 is installed. Not sure how to proceed.

I ordered dependencies in the order of discovery (docs, code, ...). Should I sort them alphabetically?

Depends on: #844 and #845


ermo (changes requested):
Please see the review here: https://github.com/AerynOS/recipes/pull/848#pullrequestreview-2865689680

We can figure out the plugdev thing later.

PR #847, opened 2025-05-17
[NFC] openssh: Fix typos in openssh-server paths

d5ng4i:


ermo (changes requested):
This needs a `just bump` and a rebuild.

PR #848, opened 2025-05-17
openssh: FIDO support

d5ng4i:
This adds FIDO support to openssh, which includes adding:

- libfido2
  - libcbor
  - hidapi

I've also fixed typos in openssh's `sysusers.conf`, and `tmpfiles.conf`.

hidapi is declared to be licensed under `GPL-3.0-or-later OR BSD-3-Clause` (dropping non OSI approved license).


ermo (changes requested):
Please roll up the dependencies for this into this PR with one commit per dependency (and close the respective PRs when you have done so).

Also, would you mind rebasing this onto the [2025-05-repo-rebuild](https://github.com/AerynOS/recipes/tree/2025-05-repo-rebuild) branch so we can get it in with the new infra?

ermo (changes requested):
@d5ng4i My take is that the BSD license is the best fit for AerynOS's packages.

Also, please squash the config file typo commit into the openssh commit proper, then rebuild openssh. If the manifest does not record the change, that is a bug.

The change will still be visible, but this is the nicest way of doing it.

PR #854, opened 2025-05-21
thunderbird: Update to v138.0.2

ReillyBrogan:
Other changes:
- Add langpacks (use the Solus source for these)
- Use libc++ instead of libstdc++
- Use system icu and system libevent
- Disable X11 support
- Update desktop file
- Clean up builddeps


hphilm (approved):
LGTM

PR #856, opened 2025-05-22
vte: Hide most useful desktop entry

nisel11:
I propose to hide the single most important .desktop entry in the entire OS, more vital than the kernel, init, and possibly oxygen itself, clearly indispensable for proper kernel functioning, but best marked with NoDisplay=true to protect innocent users from its overwhelming power.


ReillyBrogan:
Superseded by https://github.com/AerynOS/recipes/commit/ecbc68325d7e85f07367d32e295ccc2c8b6b4a48

PR #871, opened 2025-05-31
teamspeak6: Add at 6.0.0

sammypanda:
# Does not run

- It's missing libcef.so, it has a bundled version in the tar--im not sure if it's okay to use that (or how to use that).

<br>

Other:
- Couldn't see CPE (security) listing for teamspeak 6, they seem largely teamspeak 3
- Couldn't see release-monitoring listing for teamspeak 6, also just teamspeak 3


ermo:
Converting to draft because of the "Does not run" header in the PR cover letter.

malfisya:
Can we even legally redistribute it? This is a propietary software afterall. Over on Solus, we did have permission to redistribute TeamSpeak 3 client but it doesn't cover subsequent client version. Just want be careful here, cheers!

sammypanda:
oh, I have no idea. Probably not then :3 @malfisya the confusing license stuff honestly makes me not keen on it anyway. I might close rather than trying to resolve it.

PR #957, opened 2025-07-27
python-virtualenv: Add at 20.32.0

K1ngfish3r:
python-virtualenv: Add at 20.32.0

ensurepip needed to have bundled wheels for venv to be used as a module.

if this is not acceptable, i can remove this change

### DIFF
--- a/p/python-virtualenv/stone.yaml
+++ b/p/python-virtualenv/stone.yaml
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: © 2020-2025 Aeryn OS Developers
+# SPDX-FileCopyrightText: © 2020-2025 Serpent OS Developers
 #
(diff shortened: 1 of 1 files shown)


p/python-virtualenv/stone.yaml
```diff
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: © 2020-2025 Aeryn OS Developers
+# SPDX-FileCopyrightText: © 2020-2025 Serpent OS Developers
 #
```
ermo, line 3:
```suggestion
# SPDX-FileCopyrightText: © 2025-     Aeryn OS Developers
#
```

ermo (changes requested):
We really need to fix this in the boulder template.

PR #962, opened 2025-07-30
meson: add patch to add destconfdir and change upstream to commit

JustSoup312:
Adds a flag to define the destconfdir in packaging. This is an important part of getting statelessness support into upstreams.

The GitHub release of Meson is slightly different from the tag itself, so switching to the top commit of the tag fixes the patch.

Collaboration with @ermo.

This is my first time handling stone packaging, so if I messed anything up, please tell me.


ermo:
@JustSoup312 The patch as-is is technically correct.

However, our custom is to create a ./pkg dir in the recipe folder next to the stone, save and commit the patch there and then reference it as `%(pkgdir)/<the patch>`

You can then add the URI to the patch in a comment over the patch itself along with whatever commentary seems reasonable, which makes it easy for the next committer to update the patch from the source URI given on a meson update.

So something like this:
```yaml
    setup       : |
        # hermetic-usr patch adding the 'distconfdir' build-time configuration variable as an alternative to 'sysconfdir'
        # upstream: https://gitlab.postmarketos.org/JustSoup321/hermetic-patches/-/raw/trunk/build-system-patches/meson/meson-1.8.2-distconfdir.patch
        %patch %(pkgdir)/meson-1.8.2-distconfdir.patch
```

This ensures that even if we move our tree elsewhere (or you do), this version of meson will continue to build, as long as the upstream tarball is available.

ReillyBrogan:
Is there a PR for getting this into upstream meson? This honestly seems pointless to me without that, why would meson-using projects bother merging our distconfdir patches if support for it wasn't available in other distros meson packages?

ReillyBrogan:
> @JustSoup312 The patch as-is is technically correct.
> 
> However, our custom is to create a ./pkg dir in the recipe folder next to the stone, save and commit the patch there and then reference it as `%(pkgdir)/<the patch>`
> 
> You can then add the URI to the patch in a comment over the patch itself along with whatever commentary seems reasonable, which makes it easy for the next committer to update the patch from the source URI given on a meson update.
> 
> So something like this:
> 
> ```yaml
>     setup       : |
>         # hermetic-usr patch adding the 'distconfdir' build-time configuration variable as an alternative to 'sysconfdir'
>         # upstream: https://gitlab.postmarketos.org/JustSoup321/hermetic-patches/-/raw/trunk/build-system-patches/meson/meson-1.8.2-distconfdir.patch
>         %patch %(pkgdir)/meson-%(version)-distconfdir.patch
> ```
> 
> This ensures that even if we move our tree elsewhere (or you do), this version of meson will continue to build, as long as the upstream tarball is available.

It would be very unusual to rename the patch after every meson update. Just name it `distconfdir.patch` or the PR ID of the upstream PR it represents (so `upstream-10001.patch` if the PR was ID 10001).

JustSoup312:
> Is there a PR for getting this into upstream meson? This honestly seems pointless to me without that, why would meson-using projects bother merging our distconfdir patches if support for it wasn't available in other distros meson packages?

@ermo believed it best to do a patch first but I would prefer getting the patch into meson. I just don't know if they will accept it without a usecase

PR #986, opened 2025-08-15
gnome-text-editor: Bump for appstream-generator debugging

joebonrichie:
appstream-generator thinks the gnome svg icon file is empty, building the package locally we no longer have this problem. Let's test it.


ermo (approved):
"Fore!"

PR #987, opened 2025-08-17
boulder: Update to v0.25.6+git.edc5a6d

joebonrichie:
- .stone layout entries are now recorded deterministically probably fixing a large amount of reproducibility issues.


ermo (approved):
LFG!

PR #1008, opened 2025-09-02
gumbo-parser: Add at v0.13.2

bhh32:
# Reason for PR
This library is needed to build the new cosmic-reader COSMIC suite application. One of its dependencies depends on this library. I'm using the version listed in [release-monitoring.org](https://release-monitoring.org/project/214990/) due to the fact the original project listed in GitHub has been abandoned. The release-monitoring.org version is being actively updated.


g/gumbo-parser/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: 214990
+  rss: ~
```
ermo, line 3:
```suggestion
  rss: https://codeberg.org/gumbo-parser/gumbo-parser/tags.atom
```

ermo (changes requested):
Looks like codeberg also supports atom feeds for tags.

g/gumbo-parser/monitoring.yaml
bhh32: @ermo I have added the rss to the monitoring.yaml file.

g/gumbo-parser/stone.yaml
```diff
@@ -0,0 +1,24 @@
...
+upstreams   :
+    - https://codeberg.org/gumbo-parser/gumbo-parser/archive/0.13.2.tar.gz : dbdc159dc8e5c6f3f254e50bce689dd9e439064ff06c165d5653410a5714ab66
+summary     : HTML parser library implemented in C99
+description : |
+    HTML parser library implemented in C99
+license     : Apache-2.0
+builddeps   :
+    - pkgconfig(gtest_main)
+setup       : |
+    %meson
```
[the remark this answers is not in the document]
bhh32: Forgot to push the change. Now it's completed.

ReillyBrogan (changes requested):
So generally when we add new dependencies for packages we like to add them in the same PR as whatever needs them. So really this PR should be `cosmic-reader: Add at v0.0.0` (fill in the version) and should include both gumbo-parser as well as any other needed dependencies along with cosmic-reader itself. That way we don't have dependencies in the repo that are unused, the dependencies will be added alongside the packages that need them.

Does that make sense?

bhh32:
@ermo, no I'm going to roll it into the cosmic-reader PR when I get it to build. I'm more focused on the main cosmic packages right now.

PR #1009, opened 2025-09-02
More Gnome apps

malfisya:
**Summary**

- Add showtime
- Add gnome-contacts
- Modify gnome-desktop-defaults
- Add more packages to pkgset-aeryn-gnome
- Adjust gnome-mimeapps.list


ermo:
I, for one, like where this is going.

malfisya:
Notes: 
- @NomadicCore asks for Apostrophe and Fragments

f/folks/manifest.x86_64.jsonc
```diff
@@ -0,0 +1,148 @@
...
+			],
+			"files": [
+				"/usr/bin/folks-import",
+				"/usr/bin/folks-inspect",
+				"/usr/lib/folks/26/backends/bluez/bluez.so",
+				"/usr/lib/folks/26/backends/dummy/dummy.so",
+				"/usr/lib/folks/26/backends/eds/eds.so",
+				"/usr/lib/folks/26/backends/key-file/key-file.so",
+				"/usr/lib/folks/26/backends/ofono/ofono.so",
+				"/usr/lib/folks/26/backends/telepathy/telepathy.so",
```
ReillyBrogan, line 49: I think the telepathy backend and tools can be split out to subpackages like how Fedora does so.

g/gnome-contacts/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: 204994
+  rss: ~
```
ReillyBrogan, line 3: https://gitlab.gnome.org/GNOME/gnome-contacts/-/tags?format=atom

p/python-dbusmock/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: ~ # https://release-monitoring.org/ and use the numeric id in the url of project
```
ReillyBrogan, line 2: https://release-monitoring.org/project/19778/

p/python-dbusmock/stone.yaml
```diff
@@ -0,0 +1,26 @@
...
+description : |
+    With this program/Python library you can easily create mock objects on D-Bus. This is useful for writing tests for software which talks to D-Bus services such as upower, systemd, logind, gnome-session or others, and it is hard (or impossible without root privileges) to set the state of the real services to what you expect in your tests.
+license     : LGPL-3.0-or-later
+builddeps   :
+    - pkgconfig(python3)
+    - python-build
+    - python-installer
+    - python-packaging
+    - python-setuptools_scm
+    - python-wheel
```
ReillyBrogan, line 22: These should all be defined as `python(x)` deps. And several of them are already pulled in via macro.

p/pkgset-aeryn-gnome/stone.yaml
```diff
@@ -43,17 +44,22 @@ packages    :
...
             - binary(gnome-clocks)
+            - binary(gnome-contacts)
             - binary(gnome-disks)
+            - binary(gnome-font-viewer)
             - binary(gnome-tweaks)
             - binary(gnome-weather)
             - binary(loupe)
+            - binary(org.gnome.Decibels)
             - binary(papers)
+            - binary(showtime)
```
ReillyBrogan, line 61: Probably need to update the mimeapps.list in gnome-session too

s/showtime/stone.yaml
```diff
@@ -0,0 +1,30 @@
...
+    Play your favorite movies and video files without hassle. Showtime features simple playback controls that fade out of your way when you're watching, fullscreen, adjustable playback speed, multiple language and subtitle tracks, and screenshots — everything you need for a straightforward viewing experience.
+license     : GPL-3.0-or-later
+builddeps   :
+    - binary(msgfmt)
+    - binary(update-desktop-database)
+    - pkgconfig(blueprint-compiler)
+    - pkgconfig(gtk4)
+    - pkgconfig(libadwaita-1)
+rundeps     :
+    - gst-plugins-rs
```
ReillyBrogan, line 23: I don't think this is right. It's correct on Solus because we don't break up the gst-plugins-rs package, but on Aeryn we split out the gtk4 gstreamer plugin which is the only one this package actually depends on. Also since this is a python program the autodep isn't picking up any of the other required deps so it needs those specified as well.

Something like this I imagine:
```
rundeps:
  - binary(python3
  - gstreamer-plugin-gtk4
  - gtk-4
  - libadwaita
```
And then we probably want it to pull in `gstreamer-plugins-good` and `gstreamer-plugins-bad` at the very least so that it gets some codec support.

t/telepathy-glib/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: 6062
+  rss: ~
```
ReillyBrogan, line 3: https://gitlab.freedesktop.org/telepathy/telepathy-glib/-/tags?format=atom

f/folks/stone.yaml
```diff
@@ -0,0 +1,41 @@
+#
+# SPDX-FileCopyrightText: © 2020-2025 Serpent OS Developers
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : folks
+version     : 0.15.9
+release     : 1
+homepage    : https://gitlab.gnome.org/GNOME/folks/-/wikis/Home
```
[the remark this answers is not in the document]
malfisya: Wiki gnome is (going to be) dead. So, not going to change this.

PR #1041, opened 2025-09-10
Add openconnect at 9.12

CookieSource:
This Pull request adds Openconnect at version 9.12 including it's network handling upstream script.
Openconnect is used to connect to most VPN protocols. 

Reference issue #353


ermo (changes requested):
One small nit.

ermo:
@CookieSource Ping? :ping_pong:

PR #1053, opened 2025-09-14
sddm: Fix password-less autologin with gnome-keyring

hphilm:


hphilm:
A similar suggestion was made here: https://github.com/canonical/lightdm/issues/178#issuecomment-1134898885

hphilm:
Ah, may have to check on password less login: https://wiki.archlinux.org/title/SDDM#Passwordless_login

PR #1060, opened 2025-09-17
Resolve ImageMagick format recognition issue

CookieSource:
This PR adds the lastrip option and sets it to false. This is required for Imagemagick to function  as it requires the .la files to function. See also https://bugzilla.opensuse.org/show_bug.cgi?id=579798


joebonrichie:
nit: can you rename the commit with the package name prefix e.g. imagemagick: blah blah blah.

i/imagemagick/manifest.x86_64.jsonc
```diff
@@ -51,6 +51,9 @@
 				"/usr/bin/mogrify",
 				"/usr/bin/montage",
 				"/usr/bin/stream",
+				"/usr/lib/libMagick++-7.Q16HDRI.la",
```
ReillyBrogan, line 54: We only need the la files in the modules directory, not the ones in libdir. Can you please add a `rm -v %(installroot)%(libdir)/*.la` to the install section?

ReillyBrogan: And anyway even if we didn't delete them here we would have wanted them patterned into the same package the libs are in.

CookieSource:
Yep I'll look at both today.

PR #1067, opened 2025-09-25
vscodium: Add at v1.105.17075

AnonAlly:


AnonAlly:
@ermo I took the liberty of adding it because some people see vscodium as a way to use vscode/a similar interface without sending a mass of data to Microsoft. Codium is vscode without the analytics. I see that it does seem to be a duplicate, but that's why I added it.

AnonAlly:
@ermo ~~For duplicated files, how can I fix it ?~~

PR #1092, opened 2025-10-14
gamescope-session-steam: fix startup

hphilm:
- for some reason the flag -steamos3 blocks a session start


ermo (approved):
The sed thing is technically a recipe smell, but I can see how this is the most convenient way of doing it for now.

PR #1106, opened 2025-10-21
Update COSMIC DE to Beta 2

bhh32:
Update all COSMIC DE packages to Beta 2.


ermo (commented):
In general, it is nice to see that there does not appear to have been a lot of changes to the packaging side.

That said, the kind thing to do would be to spend a little time explaining why the patches we apply are needed, whether they are upstreamable, and if so, which issue/PR we are waiting for upstream to ack and merge? Think DEP-3 like.

The context is that we need to ensure that this stack is maintainable by multiple packagers who might not all be available at the same time, meaning that they will benefit greatly from async notes explaining non-obvious context.

c/cosmic-comp/pkg/0001-Allow-libdisplay-info-0.3.patch
```diff
@@ -1,38 +1,40 @@
-From 11397eb254fddd7ed5a88a02cb19da84c1cc46ff Mon Sep 17 00:00:00 2001
+From 2f002efc600186166ee1578ff8753a7495f5a6ea Mon Sep 17 00:00:00 2001
 From: Bryan Hyland <bryan.hyland32@gmail.com>
 Date: Wed, 17 Sep 2025 12:17:43 -0700
 Subject: [PATCH] Update libdisplay-info patch
```
ermo: Either the patch or the recipe (where it references the patch) could benefit from an explanation along the lines of:

> Because our version of libdisplay-info is newer than what's supported by the upstream crate, we need to patch this a little. This is not upstreamable as is.

ReillyBrogan (approved):

PR #1121, opened 2025-10-28
Update Cosmic to Beta 3

AnonAlly:


c/cosmic-comp/pkg/0001-Allow-libdisplay-info-0.3.patch
AnonAlly: I removed this patch because upstream seems to have fixed it.

PR #1124, opened 2025-10-29
Add nextcloud client 4.0.1

### LINKED ISSUE #1080: [Package Request]: Nextcloud-client
### Existing package request?

### Homepage of package

https://nextcloud.com/clients

The Nextcloud Desktop Client is a tool to synchronize files from Nextcloud Server with your computer.
We also need the deps  libp11 (PKCS#11 wrapper library )

### Licence of package?

GPL-2.0-or-later

https://github.com/nextcloud/desktop/archive/refs/tags/v3.17.2.tar.gz

Astate:
Desktop sync client for Nextcloud

Resolves #1080


n/nextcloud-client/stone.yaml
```diff
@@ -0,0 +1,39 @@
+name: nextcloud-client
+version: "4.0.1"
+release: 2
```
ermo: Needs to be 1 when we are adding a new recipe.

If this is the newest commit in your tree, you can edit the stone.yaml file to contain release: 1 (instead of 2), run `just build`, git add the changed files, and then run `git commit --amend` and then finally `git push -f`.

You are almost done.

PR #1138, opened 2025-11-05
Update kernels to 6.17.7

hphilm:
Update both kernels to 6.17.7 and rebuild NVIDIA drivers


hphilm:
Cos they are outdated and mostly for the ALLY Handheld series. Current patch-set is found in the **linux-handheld** packages.

PR #1144, opened 2025-11-08
font-noto-cjk: fix, use OTF, split into sans/serif packages, and add fontconfig

### LINKED ISSUE #1134: [To Fix] font-noto-cjk doesn't include noto sans cjk
For somewhat reasons noto sans cjk fonts are currently missing in font-noto-cjk (there're only serif ones)

BarryLhm:
Added missing sans fonts: font-noto-sans-cjk.
Moved original serif-only recipe to font-noto-serif-cjk.
Made original recipe a metapackage that pulls in two above

This closes #1134.


f/font-noto-serif-cjk/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: https://release-monitoring.org/378624
```
ermo:
```suggestion
  id: 378624
```

ermo (changes requested):
We have a standard way to do this that adds %(installroot)%(datadir)/%(name)/README file.

Hard for you to know of this specific approach in advance though. 😅

f/font-noto-sans-cjk/tmp/.moss-lockfile
ermo: Why is this file included...?

ermo:
I am not sure if I expressed myself unclearly when I mentioned that we have a standard way of deprecating, but I only meant that we always do it via a `/usr/share/(the package)/README` file in the package recipe.

We also still make a note in the recipes tree root in the `removals.md` file, so that we can handle things once moss actually _supports_ removals and replaces/obsoletes.

I do not have a strong opinion on whether font-noto-cjk should be removed XOR whether it should simply be a rundep-only package going forward.

BarryLhm:
maybe just make it a meta/virtual package, as sans and serif fonts could have been in the same recipe if there's no version diff between them

BarryLhm:
this way i could make this originally 'obsoleted' package more meaningful

ermo:
Having thought about this a bit more, why not just update this to be a single package with two sub-packages?

Then the main package can simply rundep on the sub-packages?

Then you only have one version to update and the sub-packages get created each time?

PR #1147, opened 2025-11-10
Add flatpak-builder

bhh32:
# Purpose
Flatpak-builder is the official tool to create flatpaks from source.

## Reason for adding
This will give developers who are testing the ability to create flatpaks from what they are developing on AerynOS.

## Dependencies
- **appstreamcli compose module**: The dependency is needed to build `flatpak-builder`. To get this dependency the appstream recipe needed to have the line `-Dcompose=true \` added to it.
- **debugedit**: This dependency is needed for both building flatpak-builder and building flatpaks using it.

## Testing
I have tested the dependencies listed above, which are also added as their own commits to the PR, as well as `flatpak-builder` itself. I tested the dependencies and `flatpak-builder` by using it to building an invoicing project that I have been working on and am planning on submitting to flathub.


a/appstream/manifest.x86_64.jsonc
```diff
@@ -27,19 +27,29 @@
...
 				"soname(libxmlb.so.2(x86_64))",
 				"soname(libzstd.so.1(x86_64))"
 			],
 			"files": [
 				"/usr/bin/appstreamcli",
+				"/usr/lib/appstream/appstreamcli-compose",
 				"/usr/lib/girepository-1.0/AppStream-1.0.typelib",
+				"/usr/lib/girepository-1.0/AppStreamCompose-1.0.typelib",
+				"/usr/lib/libappstream-compose.so.0",
+				"/usr/lib/libappstream-compose.so.1.1.1",
```
ReillyBrogan: Since appstream-compose adds new rundeps I think I'd prefer if we split it out similar to how Fedora does

d/debugedit/stone.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : debugedit
+version     : 5.2
+release     : 1
+homepage    : https://sourceware.org/pub/debugedit/5.2
+upstreams   :
+    - https://sourceware.org/pub/debugedit/5.2/debugedit-5.2.tar.xz : 705296803cc4403f38764e891b4ed38f8d8d4f8a9164bd4f86c9d4bedcac68dd
+summary     : debugedit provides programs and scripts for creating debuginfo and source file distributions, collect build-ids and rewrite source paths in DWARF data for debugging, tracing and profiling.
```
ReillyBrogan: Summary should be much shorter than this.

f/flatpak-builder/stone.yaml
```diff
@@ -0,0 +1,39 @@
...
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : flatpak-builder
+version     : 1.4.7
+release     : 1
+homepage    : https://github.com/flatpak/flatpak-builder
+upstreams   :
+    - https://github.com/flatpak/flatpak-builder/releases/download/1.4.7/flatpak-builder-1.4.7.tar.xz : fd5bc36fe3b974395f782e6c920d8955cee168f513370c32cc800b69acd980d0
+summary     : flatpak-builder is a tool for building flatpaks from sources.
```
ReillyBrogan: Maybe just `A tool for building flatpaks from source`

f/flatpak-builder/stone.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - pkgconfig(json-glib-1.0)
+    - pkgconfig(libcurl)
+    - pkgconfig(libelf)
+    - pkgconfig(glib-2.0)
+    - pkgconfig(ostree-1)
+    - pkgconfig(yaml-0.1)
+rundeps     :
+    - binary(debugedit)
+    - binary(flatpak)
+    - elfutils
```
ReillyBrogan: This isn't needed since it links against libelf

f/flatpak-builder/stone.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - git-lfs
+    - libyaml
+    - json-glib
```
ReillyBrogan: Both of these are also not needed.

f/flatpak-builder/stone.yaml
```diff
@@ -0,0 +1,39 @@
...
+networking  : yes
```
ReillyBrogan: Why does it need networking?

f/flatpak-builder/stone.yaml
```diff
@@ -0,0 +1,39 @@
...
+upstreams   :
+    - https://github.com/flatpak/flatpak-builder/releases/download/1.4.7/flatpak-builder-1.4.7.tar.xz : fd5bc36fe3b974395f782e6c920d8955cee168f513370c32cc800b69acd980d0
+summary     : flatpak-builder is a tool for building flatpaks from sources.
+description : |
+    flatpak-builder is a tool for building flatpaks from sources.
+license     : LGPL-2.1-only
+builddeps   :
+    - binary(appstreamcli)
+    - binary(debugedit)
+    - cmake(libxml2)
```
ReillyBrogan: For sheer consistency I'd rather this be `pkgconfig(libxml2)` instead. I generally don't use `cmake(*)` unless the package I'm building uses cmake itself. Meson uses pkgconfig for finding deps so it's better to express the builddep in the same way.

PR #1149, opened 2025-11-10
xeyes: Add at v1.3.1

BarryLhm:
add this for x client detection


ReillyBrogan (changes requested):
TBH I'm not sure we actually need this since we don't support x11 sessions. I know this is usable with xwayland apps, but are there really enough of those for this to matter? Do you have a specific use-case in mind?

PR #1153, opened 2025-11-11
reaper: Add at 0.1.0

hphilm:


r/reaper/stone.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - https://github.com/playtron-os/reaper/archive/refs/tags/0.1.0.tar.gz : 4581859aaa5c6f01ec3ef27d4beaa7769bda26ca64a4e8b038c7082ae75b98ff
+summary     : A command to monitor and annotate a process tree
+description : |
+    A command to monitor and annotate a process tree
+license     : MIT
+networking  : yes
+builddeps   :
+    - binary(meson)
+build       : |
+    %make build
```
ReillyBrogan: Literally the only thing the makefile does is call out to meson, so just use the `%meson`, `%meson_build`, and `%meson_install` macros please.

PR #1155, opened 2025-11-11
dotnet-sdk-8.0: Add at 8.0.415

hphilm:


d/dotnet-sdk-8.0/stone.yaml
```diff
@@ -0,0 +1,25 @@
...
+description : |
+    The .NET Core SDK
+builddeps   :
+    - binary(tar)
+install     : |
+    _discord_root=%(libdir)/discord
+    _dotnet_install=%(installroot)/usr/share/dotnet
+
+    %install_dir $_dotnet_install
+    tar zxfv %(sourcedir)/dotnet-sdk-8.0.415-linux-x64.tar.gz -C %(installroot)/usr/share/dotnet
```
ReillyBrogan: If we're going to add dotnet to the repo we need to build it ourselves. It's fine to use the downloaded binaries for bootstrapping the initial build only, but not like this. The initial build of dotnet will need to actually build dotnet and can use whatever sources it needs to accomplish that.

PR #1158, opened 2025-11-12
Add Dependencies for dotnet-sdk-8.0

hphilm:
When installing it as a package it seems that some dependencies were missing. Adding them now. Needed for #1155


BarryLhm:
@ermo this is also needed by dotnet 10 sdk

ReillyBrogan:
I merged numactl separately because I wanted to split out the libs and use it for perf. I don't think we need lttng-ust2.12 because as far as I can tell it can use the most recent releases of lttng-ust if we build it ourselves.

PR #1181, opened 2025-11-21
ncspot: new package

Astate:
An ncurses Spotify client written in Rust using librespot


ermo (commented):
Nit: Builddeps are not in order and the CI is complaining about it.

Are you proposiing this because:

- It is necessary for your development workflow, and
- You intend to maintain it?

If so, no problem. 😉

ReillyBrogan:
Versions should not have `v` in the string

PR #1182, opened 2025-11-22
vscodium: Update to 1.106.27818

AnonAlly:


ermo (approved):
Looks bening. Thanks!

PR #1184, opened 2025-11-24
kernels: Update to 6.17.9

hphilm:


ermo (approved):
I applaud your decision to split out the patches for linux-handheld as well. It will make things easier to rebase and maintain going forward.

In addition, this bump looks benign, though the handheld kernel does show a CI warning re. builddep order. If this is not a fluke, please consider fixing that as a follow-up [NFC] commit that does not bump the relevant stone.yaml release field.

PR #1185, opened 2025-11-26
linux-firmware: Update to 20251125

hphilm:


l/linux-firmware/manifest.x86_64.jsonc
```diff
@@ -40,6 +40,21 @@
...
+				"/usr/lib/firmware/sdca/1fa/1028/db4/1.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db4/2.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db4/3.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db4/4.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db4/42450005.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db9/1.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db9/2.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db9/3.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db9/4.bin.zst",
+				"/usr/lib/firmware/sdca/1fa/1028/db9/42450005.bin.zst",
```
ReillyBrogan: Any time files appear in `linux-firmware` it means we need to update the patterning. We're trying to keep that package as a pure meta-package that just depends on all other subpackages.

Looking into it it looks like these are Cirrus Logic firmware files so please add `/usr/lib/firmware/sdca/` to the patterns for `%(name)-cirrus`.

https://gitlab.com/kernel-firmware/linux-firmware/-/commit/62b60d741ff73ecd38342d0d166638fcc3670fbe

hphilm: Done ...

ReillyBrogan (approved):

PR #1186, opened 2025-11-26
This is what a KDL recipe could look like

jplatte:


jplatte:
I'm not sure, could be that you get strings with indented lines (except the first) then. Or it's the same.

ermo:
Just going to add an older YAML proposal, which I personally quite like in terms of organization.

The idea is that we align on the capabilities and properties we need in terms of structure and fields, where the below is just the starting point for doing so. Once we are happy with the YAML structure, we convert it to KDL.

By doing it this way, we enable Solus to converge to the KDL-equivalent stone.yaml version 2 format that is easy to then auto-convert to KDL at a later stage.

The format proposal below has foldable top-level sections in editors that support it. This organisation might also make it easier to have logically named structs containing the relevant top-level section values.

```yaml
# version 0 is the Solus format
# version 1 is the (current, frozen) Serpent OS stone.yaml format
# version 2 is the new AerynOS stone.yaml format starting point below
VERSION: 2
meta:
  name: nano
  version: '7.2'
  release: 163
  sources:
  - https://www.nano-editor.org/dist/v7/nano-7.2.tar.xz : 
    hash: 86f3442768bd2873cec693f83cdf80b4b444ad3cc14760b74361474fc87a4526
  homepage: https://www.nano-editor.org
  license: GPL-3.0-or-later
  tags:
  - system.base
  - editor.tui
  summary: Small, friendly text editor inspired by Pico
    description: |
        GNU nano is an easy-to-use text editor originally designed as a replacement for Pico, the ncurses-based editor from the non-free mailer package Pine (itself now available under the Apache License as Alpine).
config:
  arch:
  # This lists which architectures should be built from this recipe.
  # Will be translated into "kinds" that match our specific compiler flags, including the correct -march etc. options
  # These also acts as valid profile section anchors for custom build specifications
  # One of these will be defined as the "base" profile (see the profiles section below)
  - x86_64-v2
  - x86_64-v3x
  - x86_64-v4
  - i686
  - AArch64-rpi4
  - AArch64-rpi5
  (...)
  toolchain: clang
  # If flags is not specified, use the defaults
  flags:
    autodep: false
    libsplit: false
    devel: true
dependencies:
  build:
    - nanorc
  run:
    - nanorc
phases:
  fetch : |
    # the only phase in which there is network access (unless `networking: true` is enabled in config)
    # The intent is for this to be for invoking e.g. go and rust package management macros
    # so the rest of the build can be network-less
    # can be empty (backwards compat feature)   
  setup: |
    %patch %(pkgdir)/0001-Use-a-stateless-configuration.patch
    %reconfigure --enable-utf8 --docdir=%(docdir)/%(name)
  build: |
    %make
  install    : |
    %make_install
patterns   :
# Note that the idea is that all arches use a /usr/%(chost)/ prefix like exherbo (https://www.exherbo.org/docs/multiarch.html)
# which is also mentioned in https://gitlab.freedesktop.org/FHS/fhs-spec/-/issues/7
# For our currently supported base cases, it looks like this:
# - /usr/i686-pc-linux-gnu/{bin,lib,(...)}
# - /usr/x86_64-pc-linux-gnu/{bin,lib,(...)}
# This makes it much simpler to (cross-)compile in multiple passes and do analysis on each %(chost) build artefact tree
# ... which then makes it possible to coalesce and pattern identical copies of e.g. /usr/%(chost)/share/* 
# into a shared noarch subpkg.
- devel : /usr/%(chost)/man3/dfdlkj/
- noarch : /usr/%(chost)/share/*
profiles:
# A single profile can either fully override sections XOR copy in base sections.
# omitted sections are copied in from the base section.
# We may want to consider having special %(base_fetch), %(base_setup), %(base_build), %(base_install)
# definitions, so we can substitute in existing base script commands before or after profile-specific statements
# these would NOT contain the scriptbuilder pre-amble, but would be inserted verbatim from the base text blocks
- AArch64-rpi5 :
  dependencies:
    build:
    - (some extra rpi5-specific build dependency)
  phases:
    setup : |
      # some custom invocation relevant only to AArch64-rpi5
      (...)
      # followed by the already specified base profile setup block
      %(base_setup)
    build : |
      # some custom invocation
      (...)
    # install : |
      # omitted; uses base install instructions from the above install phase
- (...)
```

zkat:
@ReillyBrogan the closing `"""` determines the indentation of the whole text, so the two examples you gave are semantically different.

g/glib2/stone.kdl
```diff
@@ -0,0 +1,102 @@
...
+    "binary(xgettext)"
+    "pkgconfig(bash-completion)"
+    "pkgconfig(gobject-introspection-1.0)"
+    "pkgconfig32(libelf)"
+    "pkgconfig32(libffi)"
+    "pkgconfig32(libpcre2-8)"
+    "pkgconfig32(mount)"
+    "pkgconfig32(zlib)"
+    "python-packaging"
+}
```
zkat: I think this is a really great opportunity to reevaluate syntax:

```suggestion
build-deps {
    binary xgettext
    pkgconfig bash-completion
    pkgconfig gobject-introspection-1.0
    pkgconfig32 libelf
    pkgconfig32 libffi
    pkgconfig32 libpcre2-8
    pkgconfig32 mount
    pkgconfig32 zlib
    python-packaging
}
```

g/glib2/stone.kdl
```diff
@@ -0,0 +1,102 @@
...
+
+rundeps {
+    "binary(python3)"
+    "elfutils-libs"
+    "python-packaging"
+}
+
+environment """
+    export CFLAGS="${CFLAGS} -Wno-error=missing-include-dirs"
+    """
```
zkat: Why not have a higher-level thing than bash-style env stuff?


```suggestion
environment {
    CFLAGS "${CFLAGS} -Wno-error=missing-include-dirs"
    OTHER_ENV hello
}
```

g/glib2/stone.kdl
```diff
@@ -0,0 +1,102 @@
...
+    threads, dynamic loading, and an object system.
+    """
+
+upstream {
+    url "https://download.gnome.org/sources/glib/2.84/glib-2.84.0.tar.xz"
+    hash f8823600cb85425e2815cfad82ea20fdaa538482ab74e7293d58b3f64a5aff6a
+}
+
+license {
+    Apache-2.0
```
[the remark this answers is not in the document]
zkat: Correct. When you want to build the equivalent of `["foo", "bar", "baz"]`, you have two ways of doing that in idiomatic KDL, typically.

For short lists, you can just do: `mylist foo bar baz`.

For longer lists, you use `-` as an "anonymous node" that libraries often have support for as a convention of sorts even though it's nothing special to KDL itself, so you can list more things across lines without having to line-wrap with `\`:

```kdl
mylist {
  - foo
  - bar
  - baz
}
```

PR #1210, opened 2025-12-09
nfs-utils: Add at v2.8.4

bendik:
Seeing this might be something someone rely on.


bendik:
> Did you by any chance check if the correct kernel config options for (good, problem-free) NFS support are set...?

Mmm, good question. I only checked modprobe, and got nothing.

My guess, from what you're saying, would be that they ARE set, but if they're problem-free, how come my `mount -t nfs` commands says it doesn't know what NFS is?

bendik:
this should be all i need https://github.com/AerynOS/recipes/blob/41dcf458ceee85ccfa63c31a4608e19d7d1fe10c/l/linux-desktop/pkg/config-x86_64#L11312

But I've never built a kernel…

ermo:
@ReillyBrogan / @ermo 

Tagging us here so we can land nfs-utils as part of the next kernel update.

bendik:
> At a minimum, the `manifest*` files need to also be submitted as proof of build.

Yeah, about that. After an update of my boulder profile, I'm now seeing this trying to generate those:

```log
│P│ /tmp/script: line 56: bsdtar-static: command not found
│P│ + echo 'Failed to extract archive'
│P│ + exit 1
│P│ Failed to extract archive
Error: build: container: exited with failure: run: build recipe: failed with status code 1
```

ermo:
> > At a minimum, the `manifest*` files need to also be submitted as proof of build.
> 
> Yeah, about that. After an update of my boulder profile, I'm now seeing this trying to generate those:
> 
> ```
> │P│ /tmp/script: line 56: bsdtar-static: command not found
> │P│ + echo 'Failed to extract archive'
> │P│ + exit 1
> │P│ Failed to extract archive
> Error: build: container: exited with failure: run: build recipe: failed with status code 1
> ```

Which distribution are you on, and which command caused the failure?

bendik:
> Which distribution are you on, and which command caused the failure?

`sudo boulder build ./stone.yaml` on what seems to be an up to date aeryn.

bendik:
Did all the steps in the guide, fish shell style. I found the helpers. :)  and still getting this `bsdtar-static command not found` when calling `just build`.

Is there some cache i need to clear after having ran it with sudo before or 
something?
[log.txt](https://github.com/user-attachments/files/24101034/log.txt)

PR #1212, opened 2025-12-10
Security: Add ufw Firewall.

silverhadch:


ermo:
I am (strongly) in favour of this decision (using firewalld) it has to be said.

PR #1226, opened 2025-12-13
nfs-utils: Add at v2.8.4

bendik:
FWIW: It mounts on my machine

Also includes the
rcpbind@1.2.7


ermo:
Also, I'm doing a kernel bump. Which kernel config parameters did you need included again to properly support (the good kind of) NFS?

ermo (changes requested):
Please split rbcbind out into its own commit. Thanks.

ermo (approved):
Let's see how it goes.

n/nfs-utils/stone.yaml
```diff
@@ -0,0 +1,52 @@
...
+        --without-tcp-wrappers \
+        --enable-mountconfig \
+        --enable-ipv6 \
+        --with-systemd \
+        --enable-junction
+build       : |
+    %make
+install     : |
+    %make_install
+    mv /mason/install/sbin/* /mason/install/usr/sbin/
```
ermo: This should be %(installroot) instead of /mason/install. Sorry that I missed this.

ermo (changes requested):

PR #1227, opened 2025-12-13
neofetch: remove from repo, no longer receiving updates

raphaelreichart:
Removing neofetch from the repository because the upstream project is no longer maintained and was archived on 2024-04-26.
Source: [neofetch gh](https://github.com/dylanaraps/neofetch)


ermo:
I don't think our on-disk moss-format allows for marking packages for removal (= automatically removing them from the system if they are installed and optionally replacing them with something with equivalent functionality), so we can't remove the recipe *yet*.

IIUC, the best we can do is to mark it for removal once we grow the process that lets us auto-upgrade moss to a newer on-disk moss-format, where removals and replaces *are* supported.

And yes, you just hit one of our Alpha limitations.

@ReillyBrogan Thoughts?

PR #1244, opened 2025-12-17
cosmic-greeter: Use branded fallback background

tarkah:
Uses our artwork as the "fallback" background.

This also fixes the fallback in `cosmic-greeter` since `res/background.jpg` is backed by `git-lfs` but we weren't fetching it, which resulted in a "gray" default background. You could experience this with a `systemd-homed` user w/ luks storage since `cosmic-greeter` can't load the users settings until unlocked. 

Now both a fallback renders & it's our branded default.


bhh32 (approved):
Looks good to me.

ermo (approved):
Carrying forward the stack maintainer approval from @bhh32

PR #1246, opened 2025-12-18
sudo-rs: Update to v0.2.11

### LINKED ISSUE #1083: Issue with sudo / sudo-rs in cosmic DE sessions
After a fresh install of AerynOS with the Cosmic DE, running commands as sudo in cosmic terminal fails with:

> sudo-rs: 'command': command not found

The issue seems to be tied to cosmic terminal only. Within kitty everything seems to be working normally and as expected. I haven't tested with other terminals yet.

ermo:
- build.rs was removed, so remove corresponding AerynOS-specific patch
- To offset the patch removal and to fix #1083, add `Defaults secure_path` statement to the sudoers file.

Changelog [here](https://github.com/trifectatechfoundation/sudo-rs/releases/tag/v0.2.11)


bhh32:
1. Re-installed on my dell laptop (bare metal).
2. Opened the terminal using `Super+T`.
3. Ran the automated packaging script under root (didn't try to use sudo in the case the test could be tainted).
4. Built the PR.
5. Installed the built sudo-rs.
6. Exited out of root, and tried using sudo.

Outcome: It works as expected.

1. Disabled root moss repo local.
2. Ran `moss sync -u` to uninstall PR sudo-rs.
3. Exited out of root.
4. Tried using sudo.

Outcome: Fails with the error of sudo not recognizing the commands.

PR #1267, opened 2025-12-29
heaptrack: Add at v1.5.0

tarkah:
Heaptrack is an excellent tool for investigating memory usage for application memory profiling purposes.

I intend to maintain this tool going forward.


h/heaptrack/stone.yaml
```diff
@@ -0,0 +1,62 @@
+#
+# SPDX-FileCopyrightText: © 2025- AerynOS Developers
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : heaptrack
+version     : v1.5.0
```
ermo: Please remove the `v` or update your local boulder to the newest version, which will throw an error if you try to build this.

tarkah: fixed

PR #1278, opened 2026-01-03
Delete LICENSES directory

CookieSource:
Since we've merged the License file into the root directory this directory can be removed


ermo (approved):
I don't know if this is the right approach, but if it makes it simpler for contributors (for now), why not.

PR #1283, opened 2026-01-05
foot: Add at v1.25.0

bendik:
Added two other packages to enable building this:
* `fcft`
* `tllist`


f/foot/stone.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - pkgconfig(tllist)
+    - pkgconfig(wayland-client)
+    - pkgconfig(wayland-cursor)
+    - pkgconfig(wayland-protocols)
+    - pkgconfig(wayland-scanner)
+    - pkgconfig(xkbcommon)
+dependencies: |
+    - library(fcft)
+setup       : |
+    %meson -Dterminfo=disabled
```
bendik: Updated the pr to not include the terminfo as the terminfo for foot is already present with our current ncurses package. Should i add that as a dependency, or is there some way to resolve that conflict.

f/fcft/stone.yaml
```diff
@@ -0,0 +1,35 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+name        : fcft
+version     : 3.3.3
+release     : 1
+homepage    : https://codeberg.org/dnkl/fcft/releases/download/3.3.3
+upstreams   :
+    - https://codeberg.org/dnkl/fcft/releases/download/3.3.3/fcft-3.3.3.tar.gz : bb298772d625e7d917373d541456f34f1957a9e22b16f17f64158b2c3816563c
+summary     : A simple library for font loading and glyph rasterization using FontConfig, FreeType and pixman.
+
```
ermo: No need for a blank line here.

f/fcft/stone.yaml
```diff
@@ -0,0 +1,35 @@
...
+license     : 
+    - MIT
+builddeps   :
+    - pkgconfig(cairo)
+    - pkgconfig(check)
+    - pkgconfig(fontconfig)
+    - pkgconfig(freetype2)
+    - pkgconfig(harfbuzz)
+    - pkgconfig(librsvg-2.0)
+    # - pkgconfig(libutf8proc)
```
ermo: If these are things we do not yet have, please add a TODO comment above, and describe what it would give us to add this functionality.

f/foot/stone.yaml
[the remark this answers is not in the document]
bendik: as mentioned in my comment, if it's enabled i have a file conflict as the terminfo file for foot seems to already be bundled with our ncurses package.

ermo:
@bendik If you want this to work with terminfo, try adding `pkgconfig(ncursesw)` to builddeps for foot.

PR #1285, opened 2026-01-06
dash: add patch to work around undefined behavior

mjklbhvg:
The other, possibly better fix for this is to move the `setstackmark(&smark)` call in front of the setjmp if statement.
However, i am not comfortable with this, because i am unfamiliar with the dash codebase.
Just zero initializing might be redundant, but is the safer alternative for us.


ermo (approved):
Approved.

Ideally, this (or something like it) becomes part of upstream once the best approach has been hashed out.

PR #1291, opened 2026-01-09
brush: Add at v0.3.0

joandste:
Works exactly like our bash now but with nicer interactive.
This needs real bash for bashrc and profile. should it be rundep?


ermo:
> Works exactly like our bash now but with nicer interactive. This needs real bash for bashrc and profile. should it be rundep?

In this case, making bash a rundep of brush seems sensible on the face of it.

There is, however, a case for deliberately splitting out the stateless bash `bashrc` and `profile` files into a bash-stateless package and having both bash and brush rundep on that.

This would ensure that user systems with bash installed would still get the current stateless config, but would also ensure that brush could (at some point in the future) potentially not need the "main" bash package, on the assumption that brush achieves 100% feature parity and compatibility with "latest stable bash".

@ReillyBrogan Anything to add? Anything I might have missed in ^?

PR #1292, opened 2026-01-10
Kernels: Update to 6.18.5

hphilm:
Currently exploring some things ...


hphilm:
Ya, makes sense. Now adopted ...

hphilm:
Between 6.17 and 6.18 there were no changes in regard of ASUS patches, as you can see in handheld kernel.

ReillyBrogan (dismissed):

hphilm:
I reverted it back just to have the handheld stuff as desktop preps are done in another PR. @ermo @ReillyBrogan

hphilm:
OK, seems there was https://github.com/AerynOS/recipes/pull/1301, which is now closed, which didn't had the ASUS patches added back. Mine was just a prep for 6.18.5 without building the kernel. I can try to build mine and update the NVIDIA modules as well ...

hphilm:
gaming kernel is no problem. however building the kernel locally and then again in the CI is kinda double work. Anyway, will add the desktop later when built locally, or you merge it as is now, reuse your other PR and we can add the ASUS patches later on to the desktop kernel. Up to you ...

l/linux-handheld/pkg/depmod.conf
ermo: Are we sure this should have been deleted?

ReillyBrogan (approved):

PR #1315, opened 2026-01-18
Add dankmaterialshell & dependencies

armerpunkt:
**Summary**

Add DankMaterialShell for use with Wayland compositors like Niri, Hyprland, sway, MangoWC, and labwc.

DankMaterialShell requires quickshell, which requires CLI11

**Test Plan**

Used boulder and just to create and build. Then installed and launched it to verify it runs correctly.

**Checklist**


ermo (commented):
Generally speaking, you can see which cmake options are already set for the %cmake_* actions here:

https://github.com/AerynOS/os-tools/blob/main/boulder/data/macros/actions/cmake.yaml

If you take a closer look at that file, I am sure you will see that at least some of the options you are currently setting manually are already set correctly. 😅

ermo (dismissed):
@armerpunkt It would probably be good to create an account at https://release-monitoring.org and register DankMaterialshell there, so it gets an id.

Otherwise, this looks good to go, though if you could `git rebase -i` so the recipe + manifest changes go into the relevant commit, that would be appreciated.

armerpunkt:
Did the rebase. Hopefully that is how you wanted it. Just let me know if I need to change it.

(Also added the release-monitoring.org ID)

d/dankmaterialshell/stone.yaml
```diff
@@ -0,0 +1,49 @@
...
+    %install_file assets/danklogo.svg %(installroot)%(datadir)/icons/hicolor/scalable/apps/danklogo.svg
+    %install_file assets/systemd/dms.service %(installroot)%(libdir)/systemd/user/dms.service
+
+    %install_dir %(installroot)%(datadir)/quickshell/dms
+    cp -r quickshell/* %(installroot)%(datadir)/quickshell/dms/
+    rm -rf %(installroot)%(datadir)/quickshell/dms/.git* %(installroot)%(datadir)/quickshell/dms/.github
+
+    # Completions
+    # ./core/bin/dms completion bash > bash-completion
+    # %install_file bash-completion %(installroot)%(bashcompletionsdir)/dms
```
ermo: Is this comment (and the comments like it) necessary? You already have the correct invocation below in each case?

ermo (commented):
Not sure if I missed the commented out stuff the first time.

d/dankmaterialshell/stone.yaml
armerpunkt: Comments removed

ermo (approved):
I think this can go in now.

Don't be surprised if @ReillyBrogan makes a few small tweaks and adjustments, as he has intimate knowledge of the Qt stack etc.

armerpunkt:
Great! Thanks so much for your help in getting this merged!

PR #1319, opened 2026-01-20
Update Cosmic DE to 1.0.3

AnonAlly:
**Summary**
All is in the title

https://github.com/pop-os/repo-release/commit/7ddff8a1e2458ce18438736d7e72ec1f9e4fd723

**Test Plan**

**Checklist**


ermo:
@AnonAlly How/what did you test...?

ermo (commented):
Feels like a lot of these are only version string bumps with no upstream changes...?

Is this on purpose?

NomadicCore:
It could be because we are updating our version of cosmic between 1.0.2 to 1.0.3 that some of the packages have no changes (compared to our interim version) but now System76 is tagging to the next version, the only change might only be the string..

Externally it validates that AerynOS is on 1.0.3.

For @AnonAlly to confirm

PR #1343, opened 2026-01-25
EDK,QEMU: Add riscv64 support

Vixea:
**Summary**

Support for riscv64 binaries/OSes in QEMU and edk2 firmware

**Test Plan**

Run ubuntu 25.10 riscv64 as it requires edk2
Screenshot_2026-01-25_11-31-49

**Checklist**

- [ * ] Recipe was built and tested against the volatile stream
- [ * ] This change could gainfully be highlighted in the Stream Update notes once merged


ermo:
@ReillyBrogan I think this probably needs your input?

Vixea:
Ping?

PR #1349, opened 2026-01-27
Update GnuPG version to 2.5.17 in stone.yaml

kojq:
**Summary**

Just putting out, no merging yet. New version. Was not build and need to check deps, etc.

> GnuPG 2.5.17 and Gpg4win 5.0.1 have been released to fix a severe security bug in GnuPG versions 2.5.13 to 2.5.16. Please update as soon as possible. 

**Test Plan**

Not tested yet.

**Checklist**


ermo:
Was fixed in 70ff983d796cb8651664ea8c05f7981c046e959d

Issue #1242, opened 2025-12-17
[Package Request]: Cloudflare Warp (warp-cli)


jokokucing:
### Existing package request?

- [x] I have searched open package request reports for this package

### Homepage of package

https://developers.cloudflare.com/warp-client/get-started/linux/

### Please state why you believe this package should be included

This package is the only reason I haven't switched to AerynOS outside of a VM, I don't have the technical skills to create this package.

References:
Official Binaries : https://pkg.cloudflareclient.com/
AUR Recipe : https://aur.archlinux.org/packages/cloudflare-warp-bin
 

### Link to source archive file

https://pkg.cloudflareclient.com/pool/noble/main/c/cloudflare-warp/cloudflare-warp_2025.9.558.0_amd64.deb

jokokucing:
unfortunately I need to use Cloudflare Zero Trust for work, so I actually need the binary from Cloudflare

PR #1357, opened 2026-01-30
wgcf: Add at v2.2.30

kojq:
**Summary**

Add `wgcf` package to `recipes`.

**Test Plan**

Built as per entire `.dev` guide, ran the following, and successfully had VPN connection:

```
yes | wgcf register # or don't pipe yes if you want to read first that the project isn't affiliated with CF and CF's ToS
wgcf generate
sudo mkdir -p /etc/wireguard/
sudo cp wgcf-profile.conf /etc/wireguard/wgcf-profile.conf
# optional to change the DNS; here I changed to 1.1.1.2 as an example
sudo sed -i '/^DNS/ s/.\{1\}$/2/' /etc/wireguard/wgcf-profile.conf
# optional to clean out files if you do not intend to reference them
rm wgcf-account.toml wgcf-profile.conf
# to immediately get it up and running
systemctl enable --now wg-quick@wgcf-profile
```

**Checklist**


kojq:
linking #1242

w/wgcf/stone.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+    ./wgcf completion bash > wgcf.bash
+    ./wgcf completion zsh > wgcf.zsh
+    ./wgcf completion fish > wgcf.fish
+install     : |
+    %install_bin wgcf
+
+    %install_file wgcf.bash %(installroot)%(datadir)/bash-completion/completions/wgcf
+    %install_file wgcf.zsh %(installroot)%(datadir)/zsh/site-functions/_wgcf
+    %install_file wgcf.fish %(installroot)%(datadir)/fish/vendor_completions.d/wgcf.fish
```
ermo: Try executing `boulder recipe macros |grep completion` to get a feel for the definitions you should be using here.

ermo (approved):
LGTM, thanks!

PR #1364, opened 2026-01-31
mesa: script for vendoring rust crates and rm networking

K1ngfish3r:
**Summary**

- Removes `networking: true`
- Adds a script for "vendoring" rust crates by downloading an archive of only the subprojects directory of the main branch and sifting through the .wrap files ending with -rs
- fix lint warning

**Checklist**


ReillyBrogan:
I adapted your work into d42e358ff77b79fc85cec8f31f53a02f5eb9b83e. The main difference is that it's a python script now and that it downloads the source archive for the correct version and uses that. Your script grabbed the main branch and the crate versions frequently differ between that and released versions.

PR #1365, opened 2026-02-01
nvidia-open-gpu-kernel-modules: fix loading nvidia modules in initrd

armerpunkt:
**Summary**

- Adds systemd config option `ProtectSystem=no` to allow writing to `/usr` so that running `depmod` will work in the initrd.
- Moves the Nvidia firmware from `/eb-nvidia-fw` to `/eb-fw/nvidia` so that it will be loaded.
- Remove `nvidia-drm.fbdev=1` boot arg as it is no longer needed.

**Test Plan**

 `just build` -> `just mv-local` -> `sudo moss sync -u` -> reboot. Verify plymouth screen shows up on boot. And verify that everything still works after logging in.

**Checklist**


hphilm:
OK, whatever is decided, it needs to be rebased against `main` branch ...

PR #1366, opened 2026-02-01
mangowc: Update to v0.12.0

TheXnlyGhxst:
**Summary**
- Update mango to newest tagged release
- Remove some unnecessary dependencies
- Add potentially needed rundeps
- Reorder deps to, hopefully, be in the correct order
- Add foot and tofi to rundeps so the default config works ootb
- Add patch to change rofi to tofi in default config

**Test Plan**
- Build succeeds and currently using the new package locally

**Checklist**


m/mangowc/stone.yaml
```diff
@@ -4,40 +4,51 @@
...
+    - libxkbcommon
+    - pcre2
+    - pixman
+    - scenefx
+    - wayland
+    - wlroots
+toolchain   : gnu
 setup       : |
+    # Change default config to use tofi instead of rofi so I don't need to add another package.
+    %patch %(pkgdir)/config.patch
```
ermo: Maybe rename this to `use-tofi-instead-of-rofi.patch` so it is more descriptive?

TheXnlyGhxst: Will do. And I only specified gcc because DreamMaoMao, the mango author, does so in his AUR PKGBUILD. It did build and work just fine without it too though so I can remove that.

m/mangowc/stone.yaml
```diff
@@ -4,40 +4,51 @@
...
+    - glibc
+    - libdrm
+    - libxcb
+    - libxkbcommon
+    - pcre2
+    - pixman
+    - scenefx
+    - wayland
+    - wlroots
+toolchain   : gnu
```
ReillyBrogan: Why?

PR #1372, opened 2026-02-03
niri: Fix shell completion mixup

mauvray:
**Summary**

Fix the %fishcompletionsdir and %zshcompletionsdir macros mixup

**Test Plan**

Built the package and verified that the completions ended up in the correct places.

**Checklist**

**Resolves**:
https://github.com/AerynOS/recipes/issues/1371


ermo (approved):
LGTM, thanks!

PR #1376, opened 2026-02-04
p7zip: Add at 17.06

K1ngfish3r:
**Summary**

Adds p7zip, since bsdtar doesn't have support for password-protected .7z archives
This pkg is made based on the same from arch https://aur.archlinux.org/packages/p7zip-natspec

**Test Plan**

Extract password-protected .7z archives

**Checklist**


K1ngfish3r:
> I am thinking that the specific links in the recipe description should probably go into the PR description instead...

amended

p/p7zip/stone.yaml
```diff
@@ -0,0 +1,31 @@
...
+description : |
+    Command-line file archiver with high compression ratio
+license     :
+    - LGPL-2.1-or-later
+builddeps   :
+    - pkgconfig(libnatspec)
+tuning      :
+    - optimize: speed
+setup       : |
+    %patch %(pkgdir)/natspec.patch
```
ermo: This could do with a short description of why we add it.

I took a look at the Ubuntu PPA and this is the single-line comment I managed to come up with:

`Enable windows-like non-UTF8 international filename workarounds`

K1ngfish3r: hopefully can add on update

PR #1377, opened 2026-02-04
linux-desktop: Update to v6.18.9

ermo:
**Summary**

linux 6.18.x bugfix release.

Changelog [here](https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.18.9)

**Test Plan**

Build the kernel. Boot into it. Use system as normal.

**Checklist**


ermo:
@hphilm You should be able to push to this branch as well if need be.

PR #1380, opened 2026-02-06
NFC: Standardization of recipe version numbers (A - M) to remove " and '

NomadicCore:
**Summary**

Review and standardization of recipes to remove " and ' from version numbers for all recipes beginning A to M

**Test Plan**

No test plan as this is a "No Functional Change" update where release numbers are not being bumped.

**Checklist**


NomadicCore:
And here is why I should talk to you before going off and doing things... Okay, so everything should be in quotation marks. I guess I start again...

PR #1381, opened 2026-02-06
NFC: Update versions within recipe files to strings (A)

NomadicCore:
**Summary**

NFC: Update versions within recipe files to strings by adding quotation marks to all version numbers. This has been done for all recipes beginning with the letter A.

**Test Plan**

No test plan as the recipes haven't been bumped so haven't been rebuild

**Checklist**


NomadicCore:
Hopefully this is better / right this time?

PR #1385, opened 2026-02-07
[NFC] Versions: Convert version to strings for better ent parsing (C)

NomadicCore:
**Summary**

Update the version numbers in all recipes beginning with C to add quotation marks to make them strings. This will help due to deficiencies in yaml parsing where version numbers may not have been properly represented to boulder.

Also needed to add a monitoring.yaml file for ca-certificates

**Test Plan**

No test plan as the recipes haven't been bumped so haven't been rebuild

**Checklist**


NomadicCore:
I checked the log and I think I corrected the builddep order of curl, the release monitoring ID for ca-certificates (although I couldn't find a rss feed) and there was one conflict with Reilly's stack update yesterday I also had to resolve. I think this is not good to go?

ermo (dismissed):
LGTM. Merge at your leisure.

ermo:
You probably need to do an interactive rebase (`git rebase -i <the commit ref before the oldest commit that has a conflict>`). That will be an "interesting" experience. =)

ermo (approved):

ermo (approved):
The hundred and eleventeenth time is the charm?

EDIT: :weary:

PR #1397, opened 2026-02-09
inputplumber: Add systemd preset

hphilm:
**Summary**

Adding a preset so services won't get enabled by default. That happens dynamically via udev

**Test Plan**

Test it on OrangePi NEO to see if we have now a similar result regarding the service files:

Before: `Loaded: loaded (/usr/lib/systemd/system/inputplumber.service; enabled; preset: enabled) `
On Manjaro `Loaded: loaded (/usr/lib/systemd/system/inputplumber.service; enabled; preset: disabled)`

**Checklist**


hphilm:
I have it not yet tested ...

PR #1405, opened 2026-02-11
Add WPA_Supplicant as alternative

hphilm:
**Summary**

All major Linux Distributions use WPA_Supplicant as default backend for `networkmanager`. Using `iwd` as default might create some regressions in some edge cases.

SteamOS provides both backends: https://mafm.pages.igalia.com/blog//wifi-back-ends-in-steamos-36.html
With `wpa_supplicant-nm-backend-conf` we can now also simply activate `wpa_supplicant` as the used backend.

**Test Plan**

- test on OrangePi Neo via OGUI if Wifi Password prompt is now displayed. Related to: https://github.com/ShadowBlip/OpenGamepadUI/issues/515
- verify that Plasma still works similar as with `iwd` for Wifi setup

**Checklist**

### DIFF
--- a/w/wpa_supplicant/stone.yaml
+++ b/w/wpa_supplicant/stone.yaml
@@ -54,13 +54,13 @@ install     : |
     cd wpa_supplicant
     %make_install
     %install_bin eapol_test
-    %install_file wpa_supplicant.conf -t %(installroot)%(docdir)/wpa_supplicant
-    %install_file dbus/fi.w1.wpa_supplicant1.service -t %(installroot)/usr/share/dbus-1/system-services
+    %install_file wpa_supplicant.conf -t %(installroot)%(docdir)/wpa_supplicant/
+    %install_file dbus/fi.w1.wpa_supplicant1.service -t %(installroot)%(datadir)/dbus-1/system-services/
     %install_file dbus/dbus-wpa_supplicant.conf %(installroot)%(datadir)/dbus-1/system.d/wpa_supplicant.conf
-    #%install_file doc/docbook/*.5 -t %(installroot)%(datadir)/man/man5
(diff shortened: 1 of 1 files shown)


w/wpa_supplicant/stone.yaml
```diff
@@ -0,0 +1,78 @@
...
+build       : |
+    cd wpa_supplicant
+    %make
+    %make eapol_test
+    # Needs docbook-utils
+    # %make -C doc/docbook man
+install     : |
+    cd wpa_supplicant
+    %make_install
+    %install_exe eapol_test -t %(installroot)/usr/bin
```
ermo:
```suggestion
    %install_bin eapol_test
```

w/wpa_supplicant/stone.yaml
```diff
@@ -0,0 +1,78 @@
...
+    %install_file wpa_supplicant.conf -t %(installroot)/usr/share/doc/wpa_supplicant
```
ermo:
```suggestion
    %install_file wpa_supplicant.conf -t %(installroot)%(docdir)/wpa_supplicant/
```

ermo (changes requested):
Basically just packaging nits.

Be aware that we can only accept this into the repo on the proviso that you maintain it for the gaming edition, and that we potentially remove it once the OGUI thing supports the `iwd` backend.

cc @ReillyBrogan (just so you know that there is a clear plan for how to handle this going forward -- I respect and understand that you are decidedly *not* a fan of dealing with wpa_supplicant, so this is the least bad compromise I think we can make until OGUI sees more `iwd` related development)

hphilm:
@ermo addressed now. I will maintain it until OGUI will support `iwd`.

PR #1407, opened 2026-02-11
GNOME 49.4 Extras

joebonrichie:
**Summary**

Extra last minute gnome 49.4 updates

**Test Plan**

Use DE as normal

**Checklist**


ermo (approved):
LGTM, thanks!

PR #1411, opened 2026-02-12
inotify-tools: Add at v4.25.9.0

tarkah:
**Summary**

Adds inotify-tools, needed to debug issues we're seeing related to our atomic blitting strategy in moss around inodes & notify events.

**Test Plan**

Ran inotify and saw events come in


ReillyBrogan:
The static library is :frowning_face:

PR #1421, opened 2026-02-16
godot: Update to v4.6

TheXnlyGhxst:
**Summary**

Update Godot to most recent version.
Benefits:
- New modern theme
- Jolt physics as default
- Unique node IDs
- ...and a lot more.

**Test Plan**

Builds successfully -> Installs successfully -> Play around in the engine for a bit

**Checklist**


hphilm:
Godot was originally added as build dependency. However we can consider to maintain a version linked to OpenGamepadUI development and one as a standalone. 

For now I think AerynOS isn't the target yet for developers using our OS to develop ga.es on it, but can be in the future.

hphilm:
Now included via https://github.com/AerynOS/recipes/pull/1475

PR #1433, opened 2026-02-17
waybar: Update to v0.15.0

TheXnlyGhxst:
**Summary**

- Update to newest tagged release, info [here](https://github.com/Alexays/Waybar/releases/tag/0.15.0)
- Add stateless patch to install and find default config so waybar works ootb on install
- Re-enable cava as the build no longer fails with it
- Remove clang patch as the file it patched was changed and it no longer seems necessary (I could be wrong and have already replicated the patch using the new filename if it needs re-added)

**Test Plan**

- Build completes without issue -> Ensure all files install to correct location -> Check that all functionality works with new package and waybar finds the default config in all of the possible dirs

**Checklist**


w/waybar/stone.yaml
```diff
@@ -53,12 +53,9 @@ builddeps   :
...
-    %meson \
-        -Dcava=disabled
+    %patch %(pkgdir)/0001-make-stateless.patch
+    %meson
 build       : |
-    %meson_build
+    %ninja
 install     : |
-    %meson_install
+    %ninja_install
```
ermo: Unless there's a specific reason to do this (like a failure), please keep this as `%meson_install`.

PR #1442, opened 2026-02-19
opengamepadui: Add a patch for AerynOS

hphilm:
**Summary**

Fix Steam startup on AerynOS and add more options to Overlay mode

**Test Plan**

- Install on OrangePi Neo
- Check for regressions

**Checklist**


ermo:
@hphilm It would be really nice if the patches and the cover letter gained some more context/notes (related to your own investigative notes on Zulip), so future us will remember why we did what we did...?

PR #1526, opened 2026-03-21
p/postgresql: add postgresql 18.4

infinite-dev22:
**Summary**

Add postgresql 18.3 — an advanced object-relational database management system supporting an extended subset of SQL, including transactions, foreign keys, subqueries, triggers, user-defined types and functions.

Includes automatic split of `postgresql-devel` providing headers and `pg_config` for development against libpq.

**Test Plan**

- Built locally with `boulder build`
- Installed via local moss repository
- Verified `psql --version` returns `psql (PostgreSQL) 18.3`
- Verified `postgresql-devel` installs cleanly with correct headers

**Checklist**


ermo (commented):
This appears to be a fairly minimal build, with not a lot of dependencies enabled?

ermo (commented):
A few nits.

ermo (changes requested):
Please squash to one commit and fix builddep sorting per CI.

ReillyBrogan (approved):

ReillyBrogan:
So just to get this over the finish line I finished up some local changes to it. Note that the server unit files and whatnot are not currently installed, I'm not a fan of hardcoding the UID as 999 and I don't see why it couldn't be created with sysusers so a separate PR can be opened with that (and as a smaller scope should be easier to review).

infinite-dev22:
Noted

ReillyBrogan:
See https://github.com/AerynOS/recipes/commit/1665e5a8b71bf728a975c75e165057b8379607da. Also, are you sure that there's a point to socket activation? I'm not sure how to test that and it doesn't really seem like something other distros configure.

PR #1559, opened 2026-03-31
level-zero: move from intel-level-zero

K1ngfish3r:
**Summary**

level-zero: move from intel-level-zero
remade from https://github.com/AerynOS/recipes/pull/1464 & https://github.com/AerynOS/recipes/pull/1469

**Test Plan**

rebuild and see if intel-compute-runtime pulls in level-zero correctly

**Checklist**


ReillyBrogan (changes requested):
Because this package provides libraries and the like you'll need to copy the old version to the typos directory and make it a "shell" package that only ships a few doc files. Otherwise the "old" packages will still be available in the repo and could cause issues.

K1ngfish3r:
> Because this package provides libraries and the like you'll need to copy the old version to the typos directory and make it a "shell" package that only ships a few doc files. Otherwise the "old" packages will still be available in the repo and could cause issues.

i bumped up the release and added to typos

then realized that boulder no longer created intel-level-zero packages and instead, creates level-zero packages based on the manifest.json, which failed the package lint check

then I renamed the file to intel-level-zero, made it build intel-level-zero packages, added that to typos and still managed to fail on the package lint check

The typos folder is the bane of my PR, HELP

ReillyBrogan (approved):

ReillyBrogan:
Oh, the package name was correct but the directory was incorrect? Yeah this is the right move then.

PR #1571, opened 2026-04-04
yubico-authenticator: Add at 7.3.2

AnonAlly:
**Summary**

 Yubico Authenticator for Desktop (Windows, macOS and Linux). Manage your YubiKey and access one-time passwords with this full-featured companion app to the YubiKey.

**Test Plan**

Build it, install it, test it

**Checklist**


AnonAlly:
It's good for fixes. Ready for review

AnonAlly:
Zed format it on save...

EDIT : It's good, it's fixed

ReillyBrogan:
The main problem with this is that it uses pip to install libs rather than using system ones. That means the package is absolutely massive and if there are any security vulns there's no way for us to fix them.

AnonAlly:
I will see if I can do something ^^

ReillyBrogan:
I looked around at a few other distributions and it looks like Alpine has managed to figure out how to do it.

AnonAlly:
@ReillyBrogan should be good now

p/python-mss/stone.yaml
```diff
@@ -0,0 +1,20 @@
...
+version     : "10.2.0"
+release     : 1
+homepage    : https://github.com/BoboTiG/python-mss
+upstreams   :
+    - https://github.com/BoboTiG/python-mss/archive/refs/heads/main.zip : 9bd005a5e4fdd94ffeaeb6f0c2c6dc3676853979b15341462d36ed0f21a434c9
+summary     :  An ultra fast cross-platform multiple screenshots module in pure Python using ctypes. 
+description : |
+    An ultra fast cross-platform multiple screenshots module in pure Python using ctypes. 
+license     : 
+    - MIT
```
ReillyBrogan: When there's just one license you can just do
```
license     : MIT
```

p/python-mss/manifest.x86_64.jsonc
```diff
@@ -0,0 +1,85 @@
+/** Human readable report. This is not consumed by boulder */
+{
+	"manifest-version": "0.2",
+	"packages": {
+		"python-mss": {
+			"build-depends": [
+				"python(hatchling)"
+			],
+			"files": [
+				"/usr/bin/mss",
```
ReillyBrogan, line 11: Since there's a bin in this package you need to add `binary(python3)` as a rundep.

p/python-mss/stone.yaml
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : python-mss
+version     : "10.2.0"
+release     : 1
+homepage    : https://github.com/BoboTiG/python-mss
+upstreams   :
+    - https://github.com/BoboTiG/python-mss/archive/refs/heads/main.zip : 9bd005a5e4fdd94ffeaeb6f0c2c6dc3676853979b15341462d36ed0f21a434c9
```
ReillyBrogan: `https://pypi.debian.net/mss/mss-10.2.0.tar.gz`

ReillyBrogan: The way you did it would just use whatever the latest commit on main is, which is both not what "10.2.0" is and is also going to break almost immediately once they do a new commit.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
+# SPDX-FileCopyrightText: © 2025- AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : yubico-authenticator
+version     : "7.3.2"
+release     : 1
+homepage    : https://github.com/Yubico/yubioath-flutter
+upstreams   :
+    - https://github.com/Yubico/yubioath-flutter/archive/refs/tags/7.3.2.tar.gz : 6aba677eb4f2705201bdee0ca3a56711fcba94b42e679c4040c32d566c332f20
+    - https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.41.2-stable.tar.xz#flutter.tar.xz :
```
ReillyBrogan: You're mixing up Solus source syntax with Aeryn one. On Solus there's no `rename` so they add `#flutter.tar.xz` to achieve the same thing. Drop that part.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
+# SPDX-FileCopyrightText: © 2025- AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : yubico-authenticator
+version     : "7.3.2"
+release     : 1
+homepage    : https://github.com/Yubico/yubioath-flutter
```
ReillyBrogan: https://developers.yubico.com/yubioath-flutter/

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - https://github.com/Yubico/yubioath-flutter/archive/refs/tags/7.3.2.tar.gz : 6aba677eb4f2705201bdee0ca3a56711fcba94b42e679c4040c32d566c332f20
+    - https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.41.2-stable.tar.xz#flutter.tar.xz :
+        hash: 4a04f8a6152986d14fc137ffaf98106ca743c0f9ab66f1bc2f20ee84eb573e5c
+        unpack: false
+        rename: flutter.tar.xz
+summary     : Yubico Authenticator for Desktop (Windows, macOS and Linux) and Android
+description : |
+    Manage your YubiKey and access one-time passwords with this full-featured companion app to the YubiKey.
+license     :
+    - Apache-2.0
```
ReillyBrogan: See mss note.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+name        : yubico-authenticator
+version     : "7.3.2"
+release     : 1
+homepage    : https://github.com/Yubico/yubioath-flutter
+upstreams   :
+    - https://github.com/Yubico/yubioath-flutter/archive/refs/tags/7.3.2.tar.gz : 6aba677eb4f2705201bdee0ca3a56711fcba94b42e679c4040c32d566c332f20
+    - https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.41.2-stable.tar.xz#flutter.tar.xz :
+        hash: 4a04f8a6152986d14fc137ffaf98106ca743c0f9ab66f1bc2f20ee84eb573e5c
+        unpack: false
+        rename: flutter.tar.xz
```
ReillyBrogan: You can actually have boulder do the extraction for you, which allows for dropping your manual tar later.
```
upstreams   :
    - https://github.com/Yubico/yubioath-flutter/archive/refs/tags/7.3.2.tar.gz:
        hash: 6aba677eb4f2705201bdee0ca3a56711fcba94b42e679c4040c32d566c332f20
        unpackdir: yubioath

    - https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.41.2-stable.tar.xz#flutter.tar.xz :
        hash: 4a04f8a6152986d14fc137ffaf98106ca743c0f9ab66f1bc2f20ee84eb573e5c
        unpackdir: yubioath/flutter
```

ReillyBrogan: Side note, when manually doing tar you should use bsdtar-static instead, and specify it with `binary(bsdtar-static)`.

ReillyBrogan: Then change your PATH statement to
```
export PATH="$PATH:${PWD}/flutter/bin"
```

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    %patch %(pkgdir)/0002-fix-execute-authenticator-helper-on-init.patch
+
+    # Extract Flutter
+    tar -xf %(sourcedir)/flutter.tar.xz -C %(sourcedir)/
+    # Flutter detects the version of itself through git
+    git config --global --add safe.directory %(sourcedir)/flutter
+build       : |
+    cd helper
+    gpep517 build-wheel --wheel-dir dist --output-fd 1
+    cd ..
```
ReillyBrogan: Two things:
Use `pushd` and `popd` instead of cd here (makes it harder to make a mistake)

You don't need gpep517 at all. That's what alpine does and I'm not sure why but I tested it and it works fine with our `%pyproject_build` and `%pyproject_install` macros. In fact doing that allows you to drop a bunch of python builddeps since they're only needed at runtime while gpep517 seems to require them at buildtime too.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+    # Then, build the Flutter GUI
+    flutter pub get
+    flutter build linux --release
+install     : |
+
+    python3 -m installer --destdir="%(installroot)" helper/dist/*.whl
```
ReillyBrogan: Replace with
```
pushd helper
%pyproject_install
popd
```

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    # Flutter
+    export PATH="$PATH:%(sourcedir)/flutter/bin"
+setup      : |
+    %patch %(pkgdir)/0001-fix-don-t-install-bundled-helper-use-system-python-h.patch
+    %patch %(pkgdir)/0002-fix-execute-authenticator-helper-on-init.patch
+
+    # Extract Flutter
+    tar -xf %(sourcedir)/flutter.tar.xz -C %(sourcedir)/
+    # Flutter detects the version of itself through git
+    git config --global --add safe.directory %(sourcedir)/flutter
```
ReillyBrogan: This git command can be removed once you switch to having boulder extract the archives. Boulder will fix the permissions that made this necessary.

y/yubico-authenticator/stone.yaml
ReillyBrogan: After cleaning up the builddeps and removing gpep517 I was able to build it with just the following:
```
builddeps  :
    - binary(cmake)
    - binary(ninja)
    - binary(which)
    - pkgconfig(ayatana-appindicator3-0.1)
    - pkgconfig(gtk+-3.0)
    - pkgconfig(libnotify)
    - python(poetry-core)
```

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    cd ..
+
+    # Then, build the Flutter GUI
+    flutter pub get
+    flutter build linux --release
+install     : |
+
+    python3 -m installer --destdir="%(installroot)" helper/dist/*.whl
+
+    install -Dm755 helper/authenticator-helper.py %(installroot)%(bindir)/authenticator-helper
```
ReillyBrogan: Use `%install_exe` instead of `install -Dm755`

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+    _yubico_authenticator_root=%(libdir)/yubico-authenticator
+    _yubico_authenticator_install=%(installroot)/$_yubico_authenticator_root
+
+    %install_dir $_yubico_authenticator_install
+    cp -R build/linux/x64/release/bundle/* $_yubico_authenticator_install
+
+    sed -i 's|Exec="@EXEC_PATH/authenticator"|Exec=yubico-authenticator|' $_yubico_authenticator_install/linux_support/com.yubico.yubioath.desktop
+    sed -i 's|Icon=@EXEC_PATH/linux_support/com.yubico.yubioath.png|Icon=com.yubico.yubioath|' $_yubico_authenticator_install/linux_support/com.yubico.yubioath.desktop
```
ReillyBrogan: You can just patch `resources/linux/linux_support/com.yubico.yubioath.desktop` ahead of time instead of doing sed there.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+    %install_file $_yubico_authenticator_install/linux_support/com.yubico.yubioath.desktop %(installroot)/%(datadir)/applications/com.yubico.yubioath.desktop
+    %install_file $_yubico_authenticator_install/linux_support/com.yubico.yubioath.png %(installroot)/%(datadir)/icons/hicolor/128x128/apps/com.yubico.yubioath.png
```
ReillyBrogan: Can you just delete `$_yubico_authenticator_install/linux_support` after this? Is it used in the app itself?

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+    %install_dir %(installroot)%(bindir)
```
ReillyBrogan: bindir should already be created at this point.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    ln -svf $_yubico_authenticator_root/authenticator %(installroot)%(bindir)/yubico-authenticator
```
ReillyBrogan: Relative symlinks are almost alway preferable to absolute ones. You can make the `ln` command figure out what to use by doing this instead
```
ln -srvf %(installroot)$_yubico_authenticator_root/authenticator %(installroot)%(bindir)/yubico-authenticator
```

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    flutter pub get
+    flutter build linux --release
+install     : |
+
+    python3 -m installer --destdir="%(installroot)" helper/dist/*.whl
+
+    install -Dm755 helper/authenticator-helper.py %(installroot)%(bindir)/authenticator-helper
+
+    _yubico_authenticator_root=%(libdir)/yubico-authenticator
+    _yubico_authenticator_install=%(installroot)/$_yubico_authenticator_root
```
ReillyBrogan: I generally prefer that such variable declarations are moved to the top of the install step. Makes it easier to see where things are going.

Also, please use `${variable}` format instead of `$variable` format in the install phase. Makes it easier to see what's going on.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - python(poetry-core)
+    - python(pyscard)
+    - python(zxing-cpp)
+    - binutils
+    - python-setuptools_scm
+    - python-vcs-versioning
+    - tar
+rundeps    :
+    - ccid
+    - pcsclite
```
ReillyBrogan: I looked into this and it doesn't look like it actually uses pscslite directly. Instead it calls the pyscard library which then calls pcscd. Since pyscard is detected by the autodep you don't need to add it manually as a dependency, but since pyscard doesn't currently depend on pcscd you should bump that in this PR and add `sysbinary(pcscd)` as a rundep of it.

Then pcscd will be pulled in transitively and all will be well.

ReillyBrogan (changes requested):

p/python-pyscard/stone.yaml
```diff
@@ -14,6 +14,7 @@ license     : LGPL-2.1-or-later
 builddeps   :
     - binary(swig)
     - pkgconfig(libpcsclite)
+    - sysbinary(pcscd)
```
ReillyBrogan: Should be a rundep

y/yubico-authenticator/stone.yaml
ReillyBrogan: @AnonAlly Can you move both of these variable declarations to the top of the install phase and then make sure that `$_yubico_authenticator_root` is replaced by `${_yubico_authenticator_root}` throughout the recipe? I think that's the last thing I see todo.

y/yubico-authenticator/stone.yaml
```diff
@@ -0,0 +1,64 @@
...
+    _yubico_authenticator_root=%(libdir)/yubico-authenticator
+    _yubico_authenticator_install=%(installroot)/$_yubico_authenticator_root
+
+    %install_dir ${_yubico_authenticator_install}
+    cp -R build/linux/x64/release/bundle/* ${_yubico_authenticator_install}
+
+    %install_file ${_yubico_authenticator_install}/linux_support/com.yubico.yubioath.desktop %(installroot)/%(datadir)/applications/com.yubico.yubioath.desktop
+    %install_file ${_yubico_authenticator_install}/linux_support/com.yubico.yubioath.png %(installroot)/%(datadir)/icons/hicolor/128x128/apps/com.yubico.yubioath.png
+
+    ln -srvf %(installroot)$_yubico_authenticator_root/authenticator %(installroot)%(bindir)/yubico-authenticator
```
ReillyBrogan: Oh, `${_yubico_authenticator_root}` here as well.

AnonAlly:
Rebuild with new ISO, should be good

ReillyBrogan (approved):

PR #1589, opened 2026-04-12
suitesparse: Add at v7.12.2

HarshRelapse:
**Summary**
Packaged suitesparse, a library that contains a suite of sparse matrix algorithms.

**Test Plan**
- Check if it builds and the flags in  `cmake -L aerynos-builddir` are all sane
- Run `ctest CTestTestfile.cmake` inside the `aerynos-builddir` and check if all tests pass 

**Checklist**


HarshRelapse:
For the record, this will be a dependency for Eigen version v5.0.1. It is entirely optional and does not need to be included if Eigen or SuiteSparse is not needed. Currently in your repo, only OpenCV has a placeholder for Eigen support, but nothing else uses it (except for Monado, but that will stay in my repo since the inclusion would be too big for the soft freeze or generally).

ReillyBrogan:
It sounds like this is a dependency of something else. When adding such dependencies to the repo we strongly prefer that they be included in the PR containing whatever needs them. That way it can all be reviewed and merged together.

So for instance in this case if you were to open a PR for monado you'd have two commits in that branch, one adding suitesparse and the other adding monado.

HarshRelapse:
No problem, I can close this PR and create a new one for adding Eigen v5.0.1 with SuiteSparse as an extra commit if it's better. Both Eigen and SuiteSparse are dependencies for Monado. Since I am currently packaging a lot of packages, I thought I could contribute both deps so my repo can keep only the biggest packages (like Monado, Envision, and xrizer) and at the same time contribute a little bit :).

ReillyBrogan:
[This](https://monado.dev/) Monado?

HarshRelapse:
Yes

PR #1582, opened 2026-04-08
linux-{gaming,lts,stable}: Update to 6.18.22/7.0

hphilm:
**Summary**

Changelog: https://cdn.kernel.org/pub/linux/kernel/v7.x/ChangeLog-7.0

**Test Plan**

- Install on Minisforum X1-255
- Check for regressions

**Checklist**

- [x] Recipe was built and tested against the volatile stream
- [x] This change could gainfully be highlighted in the Stream Update notes once merged

PR #1591, opened 2026-04-12
os-artwork: Update to v2

NomadicCore:
**Summary**

Transition from old A based AerynOS logo to new Triquetra based AerynOS logo with both light and dark theme. Also remove the old SerpentOS logos as they should not be used anymore

**Test Plan**

Built and installed package on KDE Plasma and checked in about this system on both light and dark theme to check the new logo had applied correctly.

**Checklist**


hphilm:
Since #1582 landed before this one, we might merge it anyway and have step-by-step/partial updated artwork now, and as soon new kernels will land a full artwork update ...

NomadicCore:
This PR is for updating the logo so it will turn up in your about this system page on your DE.

There needs to be a separate PR to land it for Plymouth that will then need a kernel rebuild.

I (or if someone beats me to it) need to test on gnome and Cosmic at a minimum before we should merge it to confirm it's working in all environments.

Fwiw, the way the recipe is set up, the logo is placed in usr with an identical name so it should work on all systems as today. 

But always test before landing

hphilm:
Ya, I will check as soon as I boot up an AerynOS device on my end ... Mostly on Plasma.

NomadicCore:
Yeah, I've checked plasma as that's what I'm using and it works perfectly. 

The Plymouth bump (I haven't done a PR yet but is literally just bump release number and rebuild) after this PR lands also works on shutdown but we will wait on a kernel rebuild to confirm on boot up. Sounds like. It will work there too.

o/os-artwork/manifest.x86_64.jsonc
```diff
@@ -7,22 +7,10 @@
...
 			"files": [
-				"/usr/share/backgrounds/aerynos/austin-d-JQ9-cbC4rDE-unsplash.jpg",
-				"/usr/share/backgrounds/aerynos/marc-markstein-0kD1eHgyuPI-unsplash.jpg",
-				"/usr/share/backgrounds/aerynos/mike-uderevsky-4fqXRDI4gXQ-unsplash.jpg",
-				"/usr/share/backgrounds/aerynos/sami-matias-breilin-UZOpP-YHe9Q-unsplash.jpg",
-				"/usr/share/backgrounds/serpent-os",
-				"/usr/share/gnome-background-properties/austin.xml",
-				"/usr/share/gnome-background-properties/marc.xml",
-				"/usr/share/gnome-background-properties/mike.xml",
-				"/usr/share/gnome-background-properties/sami.xml",
```
ReillyBrogan, line 18: Why did you get rid of this?

NomadicCore: Apologies,

I think I understand what I did wrong. 

I removed line 18 

```DESTDIR=%(installroot) just install```

When I was trying to remove the SerpentOS branding. I think I have followed the just install file back to the artwork repo where I can see it's installing the artwork.

I'm back tonight night (Wednesday) but it will likely be tomorrow before I can fix it.

hphilm: I fixed it for you ...

ReillyBrogan (approved):

PR #1599, opened 2026-04-15
Taglib: Update to use utf8cpp

Vixea:
**Summary**

Adds utf8cpp and then makes taglib use it for building.

**Checklist**

- [X] This change could gainfully be highlighted in the Stream Update notes once merged


u/utf8cpp/manifest.x86_64.jsonc
```diff
@@ -0,0 +1,35 @@
+/** Human readable report. This is not consumed by boulder */
+{
+	"manifest-version": "0.2",
```
ReillyBrogan, line 3: Since this is a header-only package it should be collapsed into just `utf8cpp`, no devel package needs to be built.

ReillyBrogan (changes requested):

u/utf8cpp/manifest.x86_64.jsonc
Vixea: Ahh thanks it's been a second so I forgot the rules for those

ReillyBrogan (approved):

PR #1601, opened 2026-04-16
planify: Add at 4.19.4

AnonAlly:
**Summary**

https://useplanify.com/

**Test Plan**

Build it, install it, test it

**Checklist**


ReillyBrogan:
Can you upload the build logs for both of these packages somewhere and link those here?

AnonAlly:
With new version of webkitgtk I can't build it...

[gxml.log](https://github.com/user-attachments/files/26854684/gxml.log)
[planify.log](https://github.com/user-attachments/files/26854685/planify.log)

AnonAlly:
It can be built now ^^

[planify.log](https://github.com/user-attachments/files/27086630/planify.log)

g/gxml/stone.yaml
```diff
@@ -0,0 +1,25 @@
...
+    GXml is a XML parser and writer, providing both Vala and C object oriented API through GObject. 
+license     : 
+    - LGPL-2.1-or-later
+builddeps   :
+    - binary(valac)
+    - pkgconfig(gee-0.8)
+    - pkgconfig(gio-2.0)
+    - pkgconfig(libxml-2.0)
+setup       : |
+    %meson
```
ReillyBrogan, line 22: Missing translations, please check the output of the meson configure script more carefully in the future. Add `binary(msgfmt)` to add them. Also missing the GIR files, need `binary(g-ir-compiler)` for those. We don't have valadoc so you can ignore that dependency

p/planify/stone.yaml
```diff
@@ -0,0 +1,38 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : planify
+version     : "4.18.3"
+release     : 1
+homepage    : https://useplanify.com/
+upstreams   :
+    - git|https://github.com/alainm23/planify.git : c6ec62b146a801f8303093c85dd77028e67ca2be
```
ReillyBrogan: Don't use git source if you can help it. If this was because there was a patch for webkitgtk then you should have just exported the relevant commits as .patch files.

p/planify/stone.yaml
```diff
@@ -0,0 +1,38 @@
...
+summary     :  Task manager with Todoist, Nextcloud & CalDAV support designed for GNOME 
+description : |
+     Task manager with Todoist, Nextcloud & CalDAV support designed for GNOME 
+license     : 
+    - GPL-3.0-or-later
```
ReillyBrogan: When there's just one license you can just do
```
license     :   GPL-3.0-or-later
```

p/planify/stone.yaml
```diff
@@ -0,0 +1,38 @@
...
+builddeps   :
```
ReillyBrogan, line 14: I checked and there's a version 4.19.1 out. They dropped webkitgtk support there so you can just drop that specific dependency. I went through the meson output and you need to add the following dependencies:
- pkgconfig(glib-2.0)
- pkgconfig(icu-uc)
- pkgconfig(libedataserver-1.2)
- pkgconfig(libspelling-1)
- pkgconfig(sqlite3)

Some of these may be new to 4.19 but some of them were probably checked for in 4.18.3. Remember that on AerynOS we're trying to be more explicit with build dependencies. Even if something else were to pull in say `pkgconfig(sqlite3)` transitively if the current recipe checks for it you should still add it. That makes it easier to see where things are used and also makes builds more resilient to future changes in dependencies.

ReillyBrogan (changes requested):

ermo:
@AnonAlly Did you forget about this?

AnonAlly:
Um, of course not... a little forgotten x)

AnonAlly:
Should be good now.

ermo:
@AnonAlly I think you need to drop the gxml stone.yaml change from the planify commit? Or, more correctly, rebase it into the gxml commit (since it contains small, but nice changes)?

AnonAlly:
Should be good now.

PR #1603, opened 2026-04-16
plymouth: Rebuild for artwork

NomadicCore:
Bumped plymouth for os-artwork. Without a kernel rebuild this will only show the new logo at shutdown but the old logo will still be used at boot.

Once a new kernel is rebuilt, it will also show at boot time as well.

**Test Plan**

Bumped plytmouth, installed and rebooted to check new logo

**Checklist**


hphilm:
I'm fine with this getting in early. Kernel bumps are expected later this week or in the next one ... 7.1 merge-window most likely will close on the 26th of April. A new point release of 7.0 might be around the corner. 6.18-lts might get a new update by end of this week, most likely.

ReillyBrogan (changes requested):
Please remove `[nfc]` from the commit message and PR title. This _does_ have a functional change, the changing of artwork.

NomadicCore:
Should be done now.

ReillyBrogan (approved):

PR #1609, opened 2026-04-19
bash: Fix missing /etc/shells

Lunasciel:
**Summary**

Ship a minimal default login-shells list under `/usr/share/defaults/shells` and register a tmpfiles.d entry that links `/etc/shells` to it. This lets tools like `chsh` and `useradd` find the list of valid login shells out-of-the-box on a stateless system, without requiring manual setup.

**Test Plan**

**Checklist**


ermo:
I have a draft PR up for the package that owns `/usr/share/defaults/shells` [here](https://github.com/AerynOS/recipes/pull/1630).

From a statelessness perspective, registering a `tmpfiles.d` entry that links `/etc/shells` to `/usr/share/defaults/shell` is a bit hacky, as the statelessness approach is to patch things to work with both, where the /etc/(...) is the override to the distro/upstream-managed configuration.

That said, I am aware that this is the approach that e.g. PostmarketOS is going with for now.

PR #1610, opened 2026-04-19
libnetfilter_queue: New package

Lunasciel:
**Summary**

Userspace library providing an API to packets queued by the kernel packet filter via the NFQUEUE target. Required by netfilter-based userspace firewalls.

**Test Plan**


Lunasciel:
Withdrawing — no in-tree consumer after pulling the portmaster recipe. Will reopen alongside a real consumer.

PR #1612, opened 2026-04-19
logseq: Add at v0.10.15

Lunasciel:
**Summary**

A privacy-first, open-source platform for knowledge management and collaboration. Logseq lets you create outlines, notes, and knowledge graphs entirely in plain text files on your local device.

Packages the prebuilt upstream Electron bundle (linux-x64 zip) rather than building ClojureScript + Electron from source, to keep the recipe maintainable. Ships a launcher at `/usr/bin/logseq`, a desktop entry and an icon; `chrome-sandbox` is made setuid for Chromium's unprivileged-namespace sandbox.

**Test Plan**


AnonAlly:
Please follow theses guidelines


https://github.com/AerynOS/recipes#git-summary-requirements

ReillyBrogan:
In it's current state this can't be accepted since we're trying to avoid adding new bin packages (the current ones we have are a necessary evil). This would need to be built from source code.

What I would recommend is having a separate repo just for packages you build yourself. I have one myself at https://github.com/ReillyBrogan/aeryn-reilly-repo

ermo:
Just to chime in and back Reilly up here, we're sorry that this is the way it has to be for now, but we simply don't have the tooling to manage this in a good way yet, other than recommending that users keep recipes in private repos.

We're working as hard as we can to address this deficit, but things need to happen in a certain order, or the house of cards collapses on top of all of us.

Thank you for your understanding.

ermo:
@Lunasciel If you're using this from a private repo, please close this PR with a comment to that effect. Thank you.

Lunasciel:
Will be updated when logseq is ready.

PR #1613, opened 2026-04-19
network-manager-applet: Fix bundling nm-connection-editor with applet

Lunasciel:
**Summary**

Users on desktops without a GTK3 system tray (COSMIC, Sway, Hyprland, niri, …) often want the advanced NetworkManager connection editor without the systray applet. Split `nm-connection-editor` into its own subpackage that depends on `network-manager-applet` for shared assets (icons, translations, GConf convert data), so it can be installed independently without file conflicts.

Files belonging to the main applet (`nm-applet` binary, its desktop entry, its man page, all shared icons and translations) stay in `network-manager-applet`; only the editor-specific files move into the new subpackage.

**Test Plan**


ReillyBrogan (approved):
There were some minor changes I made when reviewing this that I figured would be easier to just do myself:
- You had nm-connection-editor depending on network-manager-applet still. Since it seemed like the point of this PR was to get rid of that dependency I fixed it by moving the locales into nm-connection-editor and making network-manager-applet depend on nm-connection-editor instead.
- I noticed that appindicator support was disabled. I think we added libayatana in the time since this package was last looked at seriously and it was never enabled here, so I enabled it.

PR #1614, opened 2026-04-19
nodejs: Update to v25.9.0

Lunasciel:
**Summary**

Major version bump from Node.js v24 to v25.9.0.

Notable changes:

- Recipe rewritten to build from upstream tarball with shared deps: brotli, cares, libuv, nghttp2, nghttp3, openssl, sqlite, zlib, zstd, system ICU.
- SONAME bumped to 141 per Node's `abi_version_registry.json`.
- Corepack is no longer bundled in Node 25 upstream, so the recipe no longer ships it.
- `NODEJS_PRIVATE_SITELIB` now uses `$(libdir)` for consistency with the x86_64-v2 build.

**Test Plan**


ReillyBrogan:
The nodejs package is setup VERY intentionally to be a virtual package that sets up symlinks to the "default" nodejs major version which is packaged in a separate named package (nodejs-22/nodejs-24/etc). This cannot be accepted for that reason alone, but addtionally nodejs-25 is not a LTS version and so we will not be accepting a package recipe for it.

PR #1615, opened 2026-04-19
openjdk-26: New package

Lunasciel:
**Summary**

Open implementation of Oracle's Java Development Kit, major version 26. Provides the JDK; the JRE is not separately packaged.

The recipe currently bootstraps from an Adoptium Temurin JDK 25 binary (`_bootstrap=1`). Once openjdk-26 is landed, a follow-up should flip `_bootstrap` to 0 and use the in-tree openjdk-25 as the boot JDK so we no longer pull a third-party binary at build time.

**Test Plan**


o/openjdk-26/monitoring.yaml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+releases:
+  id: ~
```
joebonrichie, line 5: https://release-monitoring.org/project/389296/

ReillyBrogan:
We're not currently planning on maintaining packages for non-LTS Java versions.

PR #1623, opened 2026-04-23
powerstation: Add 98-powerstation.preset

hphilm:
**Summary**

With upcoming preset usage, it is better to configure it within the package

Depends on #1651 

**Test Plan**

- install and test
- see for regressions

**Checklist**


hphilm:
Up for discussion if it would be better to have a separate `preset` here or within `systemd-preset-aeryn` package. Needs to be enabled to function properly within `opengamepadui`.

ermo (approved):
LGTM, thanks!

PR #1624, opened 2026-04-23
heroic-games-launcher: Add at 2.22.0

### LINKED ISSUE #1105: [Package Request]: Heroic Games Launcher
### Existing package request?

### Homepage of package

https://heroicgameslauncher.com/

For gaming Epic/GOG/Amazon.

### Licence of package?

_No response_

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v2.18.1/Heroic-2.18.1-linux-x64.tar.xz

hphilm:
**Summary**

A games launcher for GOG, Amazon and Epic Games for Linux, Windows and macOS. 
Changelog: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/tag/v2.22.0
Close #1105 and makes #1576 obsolete

**Test Plan**

Build it, install it, test it.

**Checklist**


hphilm:
I might need to revisit if the exec patch is needed, since we don't use a distro provided electron here ...

h/heroic-games-launcher/pkg/fix-exec-heroic-add-to-steam.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
```
ReillyBrogan, line 1: Can you please use `heroic` instead of `/usr/bin/heroic` in this patch so that it follows PATH?

h/heroic-games-launcher/stone.yaml
```diff
@@ -0,0 +1,43 @@
...
+networking  : true
+setup       : |
+    # Fix Add-to-Steam
+    %patch %(pkgdir)/fix-exec-heroic-add-to-steam.patch
+    # Fix Desktop file
+    %patch %(pkgdir)/fix-exec-heroic-desktop-file.patch
+    pnpm install
+    pnpm run download-helper-binaries
+build       : |
+    pnpm dist:linux tar.xz
```
ReillyBrogan: I checked and you can do `pnpm dist:linux dir` instead. Basically `tar.xz` builds the `dir` target and then compresses it to a tar.xz archive. Since we don't care about the archive we can just skip that part.

h/heroic-games-launcher/stone.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    %install_dir $_heroic_install
+    cp -R dist/linux-unpacked/* $_heroic_install
+
+    %install_dir %(installroot)%(bindir)
+    ln -rsv $_heroic_root/heroic %(installroot)%(bindir)/heroic
+
+    # Desktop file, icon
+    %install_file flatpak/com.heroicgameslauncher.hgl.desktop %(installroot)/%(datadir)/applications/com.heroicgameslauncher.hgl.desktop
+    %install_file flatpak/com.heroicgameslauncher.hgl.png %(installroot)%(datadir)/icons/hicolor/128x128/apps/com.heroicgameslauncher.hgl.png
```
ReillyBrogan: This 128x128 icon looks fugly on my 3840x2160 32" display. I looked in the repo and there's an SVG icon in one of the asset dirs that as far as I can tell is the exact same icon. Can you add this so it's installed to the appropriate place?
```
%install_file src/frontend/assets/heroic-icon.svg %(installroot)%(datadir)/icons/hicolor/scalable/apps/com.heroicgameslauncher.hgl.svg
```
For reference on the icons being identical:

Image

ReillyBrogan (changes requested):

ReillyBrogan (approved):

ReillyBrogan:
I would have preferred to include both the png icon _and_ the svg one just in case someone is using an environment that doesn't support svg icons but this is fine and that can be added later.

PR #1625, opened 2026-04-24
steamos-manager: Add 98-steamos-manager.preset

hphilm:
**Summary**

This needs to been enabled by default, otherwise Steam won't start in gamescope session.

Depends on #1651 

**Test Plan**

- install, test for regressions

**Checklist**


hphilm:
Up for discussion if it would be better to have a separate `preset` here or within `systemd-preset-aeryn` package. Without proper setup a steam session within gamescope won't work.

ermo (approved):
LGTM, thanks!

PR #1626, opened 2026-04-24
opengamepadui: Remove preset

hphilm:
**Summary**

Since we disable now by default, this is not needed anymore.

**Test Plan**

- install and test

**Checklist**


hphilm:
Obsolete by https://github.com/AerynOS/recipes/pull/1627

PR #1627, opened 2026-04-24
systemd-preset: Remove preset files not needed anymore

hphilm:
**Summary**

With the new default preset settings we don't need preset files disabling services anymore
Will obsolete https://github.com/AerynOS/recipes/pull/1626

**Test Plan**

**Checklist**


ReillyBrogan (dismissed):

ermo (commented):
@hphilm I see >1 release bumps? Is that on purpose?

ermo (approved):
LGTM, thanks!

PR #1629, opened 2026-04-24
nvidia-graphics-driver: Rename preset

hphilm:
**Summary**

Preset file needs numbering to be accounted

**Test Plan**

- install and test

**Checklist**


hphilm:
Up for discussion if it would be better to have a separate `preset` here or within `systemd-preset-aeryn` package.

ermo:
I set this to draft because it has been superseded here:

https://github.com/AerynOS/recipes/pull/1657/commits

When that PR is landed, this will be closed. Sorry for not communicating that more clearly.

PR #1637, opened 2026-04-24
lxcfs: Rename preset

hphilm:
**Summary**

Preset file needs numbering to be accounted

**Test Plan**

- install and test

**Checklist**


hphilm:
Closing now. We should still decide if we either remove the preset or correct it by adding a number to it.

PR #1642, opened 2026-04-26
envision: Add at v3.2.0

HarshRelapse:
**Summary**

Packaged Envision, Monado, xrizer, wayvr and xr-hardware

**Test Plan**

Testing is done (Valve Index and Vail as game on protonge-10-34); packaging is fully done, and dependencies and manifest paths have been triple-checked. However, due to upstream Mesa regressions, an extra environmental variable needs to be added to fix the DRM leasing problems with RDNA2 (in Envision -> Settings -> Advanced -> Environment Variables -> RADV_DEBUG=nodisplaydcc %command%).

Async Reprojection and more work robustly. Pavlov is still broken, but that's on Pavlov. (https://gitlab.freedesktop.org/monado/monado/-/work_items/574 and https://gitlab.freedesktop.org/mesa/mesa/-/work_items/15025)

It's done for merging

**Checklist**

Added Envision at v3.2.0, a graphical app that acts as an orchestrator to get a full Monado setup up and running with a few clicks.

Added Monado at v25.1.0, a open source XR runtime.


ReillyBrogan:
@HarshRelapse FYI I just made some cleanups/adjustments to suitesparse/eigen. Please make sure you don't force push in such a way that it overwrites those changes.

I will leave review comments for the rest of the packages shortly.

x/xr-hardware/stone.yaml
```diff
@@ -0,0 +1,17 @@
...
+upstreams   :
+    - https://gitlab.freedesktop.org/monado/utilities/xr-hardware/-/archive/1.1.1/xr-hardware-1.1.1.tar.gz : e4da17428a6f8a36e5c81685a3803b0cbce56098fdbf4d3db09bd6b18f3c4755
+summary     : Hardware description for XR devices
+description : |
+    A simple set of scripts to generate udev rules for user access to XR (VR and AR) hardware devices
+license     : BSL-1.0
+build       : |
+    %make clean_package all
+install     : |
+    %make_install RULES_DIR=/usr/lib/udev/rules.d
```
ReillyBrogan: Use `RULES_DIR=%(udevrulesdir)` instead.

x/xr-hardware/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: 242258
+  rss: ~
```
ReillyBrogan: Gitlab has RSS support for releases and tags.

x/xr-hardware/stone.yaml
```diff
@@ -0,0 +1,17 @@
...
+release     : 1
+homepage    : https://gitlab.freedesktop.org/monado/utilities/xr-hardware
+upstreams   :
+    - https://gitlab.freedesktop.org/monado/utilities/xr-hardware/-/archive/1.1.1/xr-hardware-1.1.1.tar.gz : e4da17428a6f8a36e5c81685a3803b0cbce56098fdbf4d3db09bd6b18f3c4755
+summary     : Hardware description for XR devices
+description : |
+    A simple set of scripts to generate udev rules for user access to XR (VR and AR) hardware devices
+license     : BSL-1.0
+build       : |
+    %make clean_package all
```
ReillyBrogan: The makefile has this next to the `clean_package` target:
```
# Currently does nothing, since the generated files get committed.
```

m/monado/stone.yaml
```diff
@@ -0,0 +1,72 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : monado
+version     : "25.1.0"
+release     : 1
+homepage    : https://gitlab.freedesktop.org/monado/monado
```
ReillyBrogan: `https://monado.freedesktop.org/`

m/monado/stone.yaml
```diff
@@ -0,0 +1,72 @@
...
+    # - pkgconfig(LeapSDK)
+    # - pkgconfig(LeapV2)
+    # - pkgconfig(ONNXRuntime)
+    # - pkgconfig(OpenHMD)
+    # - pkgconfig(Percetto)
+    # - pkgconfig(depthai)
+    # - pkgconfig(realsense2)
+setup       : |
+    # Fix Pavlov VR broken controller mappings: https://discord.com/channels/1065291958328758352/1093090359464165470/1495868738992341113
+    sed -i 's/XRT_MAX_HANDLE_CHILDREN 256/XRT_MAX_HANDLE_CHILDREN 512/' src/xrt/state_trackers/oxr/oxr_objects.h
```
ReillyBrogan: This should be a source code patch.

m/monado/stone.yaml
```diff
@@ -0,0 +1,72 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : monado
+version     : "25.1.0"
+release     : 1
+homepage    : https://gitlab.freedesktop.org/monado/monado
+upstreams   :
+    - git|https://gitlab.freedesktop.org/monado/monado.git : 628c83b4eb173250ed7f305b3bce6a0b41a54791
```
ReillyBrogan, line 9: Use the source archive, not the git source.

m/monado/stone.yaml
```diff
@@ -0,0 +1,72 @@
...
+summary     : The open source XR runtime
+description : |
+    Monado is an open source XR runtime delivering immersive experiences such as VR and AR on mobile, PC/desktop,
+    and any other device (because gosh darn people come up with a lot of weird hardware).
+    Monado aims to be a complete and conforming implementation of the OpenXR API made by Khronos.
+    The project is primarily developed on GNU/Linux, but also runs on Android and Windows.
+    "Monado" has no specific meaning and is just a name.
+license     : BSL-1.0
+rundeps     :
```
ReillyBrogan: I expect to see `rundeps` below `builddeps`

m/monado/stone.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - xr-hardware
+builddeps   :
```
ReillyBrogan, line 18: From the cmake output I can see that this checks for cjson which we have in the repo.

ReillyBrogan: And doxygen

x/xrizer/stone.yaml
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : xrizer
+version     : "0.5"
+release     : 1
+homepage    : https://github.com/Supreeeme/xrizer
+upstreams   :
+    - git|https://github.com/Supreeeme/xrizer.git : dd7f12103ed27359a60f0ce8c00c8db1dc22099e
```
ReillyBrogan: Use the source tarball.

x/xrizer/stone.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - binary(python3)
+    - pkgconfig(wayland-client)
+    - pkgconfig(x11)
+networking  : true
+setup       : |
+    %cargo_fetch
+build       : |
+    %cargo_build
+install     : |
+    %install_exe target/x86_64-unknown-linux-gnu/release/libxrizer.so %(installroot)%(libdir)/xrizer/bin/linux64/vrclient.so
```
ReillyBrogan: `%install_exe %(cargo_target_dir)/libxrizer.so %(installroot)%(libdir)/xrizer/bin/linux64/vrclient.so`

x/xrizer/stone.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+    %install_file /dev/null %(installroot)%(libdir)/xrizer/bin/version.txt
```
ReillyBrogan: Why an empty file?

x/xrizer/stone.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - git|https://github.com/Supreeeme/xrizer.git : dd7f12103ed27359a60f0ce8c00c8db1dc22099e
+summary     : XR-ize your favorite OpenVR games
+description : |
+    xrizer is a reimplementation of OpenVR on top of OpenXR. This enables you to run OpenVR games through any OpenXR runtime without running SteamVR.
+license     : GPL-3.0-or-later
+builddeps   :
+    - binary(cmake)
+    - binary(glslc)
+    - binary(make)
+    - binary(python3)
```
ReillyBrogan: Doesn't seem to be necessary.

x/xrizer/stone.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - pkgconfig(wayland-client)
+    - pkgconfig(x11)
+networking  : true
+setup       : |
+    %cargo_fetch
+build       : |
+    %cargo_build
```
ReillyBrogan, line 26: Change this to
```
build       : |
    export CMAKE_GENERATOR=Ninja
    %cargo_build
```
And then change `binary(make)` to `binary(ninja)`.

w/wayvr/stone.yaml
```diff
@@ -0,0 +1,42 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : wayvr
+version     : "26.2.1"
+release     : 1
+homepage    : https://wayvr.org/
+upstreams   :
+    - git|https://github.com/wayvr-org/wayvr.git : 9734955ebb60c9b3c231d8d9e64f01c6e845d77d
```
ReillyBrogan: Also needs to be source archive.

w/wayvr/stone.yaml
```diff
@@ -0,0 +1,42 @@
...
+summary     : A lightweight OpenXR/OpenVR overlay for Wayland and X11 desktops
+description : |
+    Your way to enjoy VR on Linux. Run apps inside VR & access your Wayland/X11 screens from Monado/WiVRn/SteamVR.
+license     : GPL-3.0-or-later
+builddeps   :
+    - binary(clang)
+    - binary(cmake)
+    - binary(python3)
```
ReillyBrogan: You never need to add clang unless for some reason you are building with `toolchain: gnu` (which you should almost never do). In fact none of these dependencies seem necessary to the build.

w/wayvr/stone.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - pkgconfig(x11)
+    - pkgconfig(xext)
+    - pkgconfig(xkbcommon)
+    - pkgconfig(xkbcommon-x11)
+    - pkgconfig(xrandr)
+networking  : true
+setup       : |
+    %cargo_fetch
+build       : |
+    %cargo_build --no-default-features --features=wayland,x11,openxr,pipewire,osc
```
ReillyBrogan: If the only two things you need are `wayvr` and `wayvrctl` you can explicitly only have cargo build those by using `--package wayvr --package wayvrctl`.

ReillyBrogan: I checked and default features is defined as `default = ["openvr", "openxr", "osc", "wayland", "x11"]`, so with this we're only disabling `openvr`. Is there a reason for doing that? (Hint: the reasoning behind it should be a comment in the recipe).

Also I like to see cargo feature lists ordered alphabetically.

ReillyBrogan: And finally for non-trivial build lines like this I like to see them split out to separate lines so that git blame is a bit more useful in the future for figuring out when/why something was added.

So for this I'd rather see it look like:
```
%cargo_build \
        --package wayvr \
        --package wayvrctl \
        --no-default-features \
        --features=openxr,osc,pipewire,wayland,x11
```

w/wayvr/stone.yaml
```diff
@@ -0,0 +1,42 @@
...
+install     : |
+    %cargo_install wayvr
+    %cargo_install wayvrctl
+    %install_file wayvr/wayvr.desktop %(installroot)/usr/share/applications/wayvr.desktop
+    %install_file wayvr/wayvr.png %(installroot)/usr/share/icons/hicolor/128x128/apps/wayvr.png
+    %install_file wayvr/wayvr.svg %(installroot)/usr/share/icons/hicolor/scalable/apps/wayvr.svg
```
ReillyBrogan: `/usr/share` can be replaced by `%(datadir)`

e/envision/stone.yaml
```diff
@@ -0,0 +1,38 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : envision
+version     : "3.2.0"
+release     : 1
+homepage    : https://gitlab.com/gabmus/envision
+upstreams   :
+    - git|https://gitlab.com/gabmus/envision.git : 0149a82705e97b2f9cdccbc6076b4f213c9dc890
```
ReillyBrogan, line 9: Source archive

e/envision/stone.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - pkgconfig(gio-2.0)
+    - pkgconfig(glib-2.0)
+    - pkgconfig(gtk4)
+    - pkgconfig(libadwaita-1)
+    - pkgconfig(openxr)
+    - pkgconfig(vte-2.91-gtk4)
+networking  : true
+setup       : |
+    %meson \
+      -Dprofile=release
```
ReillyBrogan, line 34: It looks like `-Dprofile=release` is only necessary because you are using git source. It can be removed after switching to the source archive.

e/envision/stone.yaml
```diff
@@ -0,0 +1,38 @@
...
+homepage    : https://gitlab.com/gabmus/envision
+upstreams   :
+    - git|https://gitlab.com/gabmus/envision.git : 0149a82705e97b2f9cdccbc6076b4f213c9dc890
+summary     : Orchestrator for the free XR stack
+description : |
+    Envision is a graphical app that acts as an orchestrator to get a full Monado setup up and running with a few clicks
+license     : AGPL-3.0-or-later
+rundeps     :
+    - monado
+    - wayvr
```
ReillyBrogan: `binary(wayvr)`

e/envision/stone.yaml
```diff
@@ -0,0 +1,38 @@
...
+release     : 1
+homepage    : https://gitlab.com/gabmus/envision
+upstreams   :
+    - git|https://gitlab.com/gabmus/envision.git : 0149a82705e97b2f9cdccbc6076b4f213c9dc890
+summary     : Orchestrator for the free XR stack
+description : |
+    Envision is a graphical app that acts as an orchestrator to get a full Monado setup up and running with a few clicks
+license     : AGPL-3.0-or-later
+rundeps     :
+    - monado
```
ReillyBrogan: Set this to whatever binary it actually calls (such as `binary(monado-cli)`)

ReillyBrogan (changes requested):

e/envision/stone.yaml
HarshRelapse: I use an git version for Monado for now since Monado 25.1.0 is broken. For Envision, system profiles are only in later commits that haven't been tagged yet, so I also use a git commit version there. For the future a source archive is ofc better, but currently it's not possible.

e/envision/stone.yaml
HarshRelapse: Will do once Envision gets a new version tag.

m/monado/stone.yaml
HarshRelapse: Adding cjson to the builddeps adds it as a rundep when installing monado. I added both now regardless.

m/monado/stone.yaml
HarshRelapse: Monado and Envision are both git, when they release stable I can switch.

HarshRelapse:
Needs retesting for new changes made. Will do later but looks sane from changes that were requested.

HarshRelapse:
Retested with Vail and protonge 10-34 on a fresh install of AerynOS. It works without any problems and is ready to merge. Pavlov and Half-Life: Alyx don't work well, but that is not a packaging problem. There are many steps to take before you have a running setup, so I would add a step-by-step guide either at https://wiki.vronlinux.org/ or somewhere else. Still, those are not a packaging error, and it's across distros like Arch so debug steps are shared across.

e/envision/stone.yaml
ReillyBrogan: Ok, if you're going to use a git ref like that then you're going to need to use a different version string to indicate that this is not actually "3.2.0". Like `3.2.0+git.0149a827`

x/xr-hardware/monitoring.yaml
ReillyBrogan: No, like `https://gitlab.freedesktop.org/monado/utilities/xr-hardware/-/tags?format=atom` or `https://gitlab.freedesktop.org/monado/utilities/xr-hardware/-/releases.atom`

x/xr-hardware/stone.yaml
ReillyBrogan: You got rid of the entire build step, but what I meant was to do `%make all` instead. Since you got rid of that the build is now happening during the install phase which is not what we want.

m/monado/stone.yaml
ReillyBrogan: See comment on envision.

m/monado/stone.yaml
ReillyBrogan: > Adding cjson to the builddeps adds it as a rundep when installing monado

Which is a good thing because we want it to use the package version of libcjson. If you didn't do that then it would build a bundled version which we can't keep track of and which would result in bigger binaries.

w/wayvr/stone.yaml
ReillyBrogan: Your comment is explaining that you disabled OpenVR support but not **why**. Why do we not want OpenVR support?

x/xr-hardware/stone.yaml
HarshRelapse: After looking over everything again, I discovered that the build only creates unnecessary .hwdb files because the udev file is the only required package file and the 70-xr-hardware file isn't affected by the build. So I removed the build step and checked the same file, and nothing changed, so it was unnecessary.

e/envision/stone.yaml
HarshRelapse: Resolved for Monado and Envision.

x/xr-hardware/monitoring.yaml
HarshRelapse: Updated all gitlab urls in all montoring.yaml to use tags if possible, and when not releases.

w/wayvr/stone.yaml
HarshRelapse: Added comment to why everything except for OpenVR is enabled with reason.

HarshRelapse:
Update and cleanup for all of the requested changes.

ReillyBrogan (approved):

PR #1644, opened 2026-04-26
swayosd: Add at v0.3.1

christian-bendiksen:
**Summary**

 A GTK based on screen display for keyboard shortcuts like caps-lock and volume 

**Test Plan**

Tested on hardware with current dotfiles

**Checklist**


s/swayosd/stone.yaml
```diff
@@ -0,0 +1,32 @@
...
+homepage    : https://github.com/ErikReider/SwayOSD
+upstreams   :
+    - https://github.com/ErikReider/SwayOSD/archive/refs/tags/v0.3.1.tar.gz : c8ba093637d21c25db7bccdb0009362aa712b173a1c403e87ab678a84dae9b32
+summary     :  A OSD window for common actions like volume, playback and capslock.
+description : |
+    A OSD window for common actions like volume, playback and capslock.
+license     : 
+    - GPL-3.0
+networking  : true
+builddeps   :
```
AnonAlly, line 16: Just need to reorder your deps ^^.

AnonAlly (dismissed):
Just a change required, otherwise it looks good to me ^^.

ReillyBrogan (changes requested):
I checked and this ships files under `/etc/xdg/swayosd` which may or may not be necessary (should probably include them anyway). The source code uses glib functions to locate them which follow the XDG_CONFIG_DIRS environmental variable so we should be able to move them to `/usr/share/xdg/swayosd` instead.

To do this please add a patch that changes this line in meson.build:
```
config_path = join_paths(get_option('sysconfdir'), 'xdg', 'swayosd')
```
To:
```
config_path = join_paths(get_option('datadir'), 'xdg', 'swayosd')
```

Also please re-order the builddeps as @AnonAlly mentioned.

christian-bendiksen:
> I checked and this ships files under `/etc/xdg/swayosd` which may or may not be necessary (should probably include them anyway). The source code uses glib functions to locate them which follow the XDG_CONFIG_DIRS environmental variable so we should be able to move them to `/usr/share/xdg/swayosd` instead.
> 
> To do this please add a patch that changes this line in meson.build:
> 
> ```
> config_path = join_paths(get_option('sysconfdir'), 'xdg', 'swayosd')
> ```
> 
> To:
> 
> ```
> config_path = join_paths(get_option('datadir'), 'xdg', 'swayosd')
> ```
> 
> Also please re-order the builddeps as @AnonAlly mentioned.

Everything seems to work fine after the patch.

s/swayosd/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
```
ReillyBrogan, line 4: File is missing the SPDX header.

ReillyBrogan (approved):

PR #1664, opened 2026-05-01
pkgset-oxidize: Add at v2026.04

christian-bendiksen:
**Summary**

Pkgset that provides the dependency set for oxidize window manager config across Niri, Mango, and Hyprland compositors.

**Test Plan**

Tested on hardware/my own setup

**Checklist**


ermo (approved):
On the surface of it, this looks neat. Let's try it!

PR #1671, opened 2026-05-02
qemu: Update qemu-guest-agent.service to remove WantedBy

kojq:
Remove specific WantedBy target from service file.

**Summary**

`dev-virtio\x2dports-org.qemu.guest_agent.0.device` was removed from `WantedBy`

**Test Plan**

I ran `sudo systemctl preset-all`. I believe hotplug should still work because we have `ENV{SYSTEMD_WANTS}=` in `99-qemu-guest-agent.rules`, so this is redundant and also introducing a warning.

However, anyone using the feature to auto-start on hotplug can confirm.

**Checklist**

### DIFF
--- a/q/qemu/pkg/qemu-guest-agent.service
+++ b/q/qemu/pkg/qemu-guest-agent.service
@@ -13,4 +13,4 @@ Restart=always
 RestartSec=0
 
 [Install]
-WantedBy=dev-virtio\x2dports-org.qemu.guest_agent.0.device
+WantedBy=


ermo:
@ReillyBrogan I think this needs your :eyes:

ermo (approved):
Let's give it a try.

ReillyBrogan:
I asked in Matrix if this had been tested on whether or not it would break qemu VMs. That question was never answered so I'm going to revert this.

ermo:
> I asked in Matrix if this had been tested on whether or not it would break qemu VMs. That question was never answered so I'm going to revert this.

Roger.

PR #1673, opened 2026-05-03
gram: Add at 1.2.1

K1ngfish3r:
**Summary**

gram: Add at 1.2.1

**Test Plan**

Defaults to Emacs keybindings. This commit was pushed using gram

**Checklist**


ermo (approved):
Sure, why not.

PR #1678, opened 2026-05-03
fcitx5: Add at 5.1.19

### LINKED ISSUE #1672: [Package Request]: fcitx5
### Existing package request?

### Homepage of package

https://github.com/fcitx/fcitx5

我没有输入法可用

### Licence of package?

LGPL-2.1-or-later AND Unicode-DFS-2016

https://github.com/fcitx/fcitx5

lumi-me-not:
**Summary**

This adds the `fcitx5`, `fcitx5-qt`, `fcitx5-gtk` and `fcitx5-configtool` packages.

Fixes https://github.com/AerynOS/recipes/issues/1672.

**Test Plan**

The `fcitx5` daemon runs and can be configured. Actually using the IME has not been tested yet.

**Checklist**


lumi-me-not:
I've added libhangul and fcitx5-hangul. Confirmed that these work, as well.

I could split them out into a separate PR, though. I could at least use this to confirm that everything is working.

CookieSource:
Which input methods did you add currently? Korean & English (default it comes with)

lumi-me-not:
> Which input methods did you add currently? Korean & English (default it comes with)

I added the Korean input via libhangul, I will check later what built-in modules there are, when I'm at my AerynOS PC.

lumi-me-not:
Looks like by default it only has modules for direct keyboard input. I can type Hangul with the fcitx5-hangul and libhangul.

lumi-me-not:
Rebased onto the latest main, added `-DCMAKE_INSTALL_SYSCONFDIR=%(datadir)` to the build script for `fcitx5` to make sure `/usr/share/xdg/Xwayland-session.d/20-fcitx-x11` and `/usr/share/xdg/autostart/org.fcitx.Fcitx5.desktop` are put in the right place.

K1ngfish3r:
tested for 5.1.21, can rename PR and make ready for review

ermo:
- Surprised to see name(gettext) being included as a builddep, considering that name(gettext-devel) pulls it in? Help me understand this choice?
- Surprised to see binary(tar) in the builddeps here? Does bsdtar-static not work because something calls plain tar in a configure/makefile?

lumi-me-not:
> * Surprised to see name(gettext) being included as a builddep, considering that name(gettext-devel) pulls it in? Help me understand this choice?
> 
> * Surprised to see binary(tar) in the builddeps here? Does bsdtar-static not work because something calls plain tar in a configure/makefile?

Will have a look at this later. I'm guessing it was just a brainfart on my end. x)

lumi-me-not:
`libhangul` uses `tar` so can't remove that dep.

Fixed the `gettext` deps, though. :)

Also rebased onto latest `main`.

f/fcitx5/stone.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - pkgconfig(xkbcommon)
+    - pkgconfig(xkbcommon-x11)
+    - pkgconfig(xkbfile)
+    - pkgconfig(xkeyboard-config)
+    - pkgconfig(zlib)
+    - gettext-devel
+setup       : |
+    rmdir third_party/yoga
+    mv /mason/build/x86_64/yoga third_party/yoga
+    %cmake -DCMAKE_INSTALL_SYSCONFDIR=%(datadir)
```
ermo, line 49: Does this imply that we need to set up a symlink via a systemd tmpfiles.d stanza from /etc/(...) to /usr/share/defaults/(...)?

K1ngfish3r: no, follows the same convention as other recipes which have `/usr/share/xdg/autostart` such as https://github.com/AerynOS/recipes/blob/24c3d3be25e6c0dfe7c15427e304d7052b3d4459/g/gnome-keyring/stone.yaml#L43-L44

ermo: @K1ngfish3r 

Please add a comment to that effect then?

ermo (approved):

PR #1679, opened 2026-05-03
imv: Add at v5.0.1

lumi-me-not:
**Summary**

This PR adds the imv image viewer: https://sr.ht/~exec64/imv

This also adds `libnsgif` and `libnsbmp` for gif and bmp support.

**Test Plan**

Built, installed and viewed some images with it.

**Checklist**


i/imv/stone.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - https://git.sr.ht/~exec64/imv/archive/v5.0.1.tar.gz : 8949c1df4b933b1d324e02ce49f1834a4b73dd25fa8103579e0ed105149e080e
+summary     : imv is a command line image viewer intended for use with tiling window managers
+description : |
+    imv is a command line image viewer intended for use with tiling window managers
+license     : 
+    - JSON
+    - MIT
+    - X11
+    - X11-distribute-modifications-variant
+    - Xnet
```
K1ngfish3r:
```suggestion
    - MIT
```

l/libnsbmp/stone.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+name        : libnsbmp
+version     : "0.1.7"
+release     : 1
+homepage    : https://www.netsurf-browser.org/projects/libnsbmp
+upstreams   :
+    - https://github.com/netsurf-browser/libnsbmp/archive/77ce1da1f484de5a7696d9bac0a8b42fa8cb35e2.tar.gz : e60edb17685b490c9abdfa921ddd9c20845954efecafb1d7b088100949f3eb18
+    - https://github.com/netsurf-browser/buildsystem/archive/0005ae300283ff01c2e2b05e7376b3e55dea21f7.zip :
+        hash      : bd01cead7b12cf01f60875e958a97ed8c06e2f18c14d472940b7177b202373eb
+        unpackdir : buildsystem
```
K1ngfish3r:
```suggestion
    - git|https://github.com/netsurf-browser/buildsystem.git :
        ref: 0005ae300283ff01c2e2b05e7376b3e55dea21f7
        clonedir: buildsystem
```

i/imv/stone.yaml
K1ngfish3r: i believe it should be MIT only

l/libnsbmp/stone.yaml
K1ngfish3r: can change most of these to use the `git:ref`

i/imv/stone.yaml
lumi-me-not: Yeah, true.

l/libnsbmp/stone.yaml
lumi-me-not: You mean only the buildsystem repo? Or the libnsgif and libnsbmp ones as well?

K1ngfish3r: yes for both, it would be best to use git ref, then change versioning to "%(version)+git.(short-commit-number)" so something like `"0.1.7+git.77ce1da"` for libnsbmp

lumi-me-not: Oki, done.

l/libnsbmp/stone.yaml
```diff
@@ -0,0 +1,22 @@
...
+        clonedir : buildsystem
+summary     : Libnsbmp is a decoding library for BMP and ICO image file formats, written in C
+description : |
+    Libnsbmp is a decoding library for BMP and ICO image file formats, written in C.
+license     : 
+    - MIT
+build       : |
+    %make NSSHARED=../buildsystem COMPONENT_TYPE=lib-shared
+install     : |
+    %make_install NSSHARED=../buildsystem PREFIX=/usr COMPONENT_TYPE=lib-shared
```
ermo: `PREFIX=%(prefix)` is probably more portable here.

l/libnsgif/stone.yaml
```diff
@@ -0,0 +1,22 @@
...
+        clonedir : buildsystem
+summary     : LibNSGIF is a C library for decoding GIF format images and animations
+description : |
+    LibNSGIF is a C library for decoding GIF format images and animations.
+license     : 
+    - MIT
+build       : |
+    %make NSSHARED=../buildsystem COMPONENT_TYPE=lib-shared
+install     : |
+    %make_install NSSHARED=../buildsystem PREFIX=/usr COMPONENT_TYPE=lib-shared
```
ermo: Another `PREFIX=%(prefix)` portability thing.

ermo (commented):
Are you sure we need to build this with X11 support...?

lumi-me-not:
> Are you sure we need to build this with X11 support...?

I've disabled X11 support.

ermo (approved):
Let's try this on for size.

PR #1682, opened 2026-05-04
steam: use steam-devices for udev

HarshRelapse:
**Summary**

Split Steam udev files into a separate steam-devices package to allow independent updates from Git and ensure support for hardware like the Nintendo Switch 2, Bigscreen Beyond, and the new Steam controller. Following the example of Fedora and Arch Linux

I also updated the patch but kept the author since I didn't change anything on the patch itself, and it's not my work at all.

Closes issue: https://github.com/AerynOS/recipes/issues/1646

**Test Plan**

Build both packages, then verify that the new udev files match those from upstream (they changed, and the patch also worked).

**Checklist**


s/steam-devices/stone.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - git|https://github.com/ValveSoftware/steam-devices.git : 39e7bd00f7a322e5165fd8f416b31d23daf6d385
+summary     : Steam and SteamVR udev rules
+description : |
+    List of devices Steam and SteamVR will want read/write permissions on
+license     : MIT
+setup       : |
+    %patch %(pkgdir)/0001-fix-udev-rules.patch
+install     : |
+    %install_dir %(installroot)%(libdir)/udev/rules.d
+    %install_file 60-steam-input.rules %(installroot)%(libdir)/udev/rules.d/.
```
ermo: We typically use `-t %(installroot)(...)/` instead of adding the `.` at the end when not changing the filename.

HarshRelapse: Done

s/steam-devices/stone.yaml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : steam-devices
+version     : "1.0.0.85"
```
ermo:
```suggestion
version     : "1.0.0.85+39e7bd0"
```

ermo (changes requested):
Missed this the first time around.

s/steam-devices/stone.yaml
HarshRelapse: Really good catch. Has been fixed "1.0.0.85+git.39e7bd00".

HarshRelapse:
Should be done for merging.

ermo (approved):
Let's try this out in practice.

PR #1690, opened 2026-05-07
[NFC] suitesparse: remove build failure workaround

HarshRelapse:
**Summary**

Since glibc got updated, the workaround is not needed anymore for a build.

**Test Plan**

Removed patch and checked if it build. Manifest stayed the same (as it should be).

**Checklist**


ermo (approved):
LGTM, thanks!

PR #1695, opened 2026-05-07
vapoursynth: Add at R77

K1ngfish3r:
**Summary**

vapoursynth: Add at R77

**Test Plan**

~~as per https://www.vapoursynth.com/doc/installation.html
must run `vapoursynth register-install` to allow programs to find VSSCRIPT~~
now ‎‧₊˚✧[stateless]✧˚₊‧

**Checklist**


K1ngfish3r:
R76C1 is out, im expecting a few more releases within this period

ermo:
@ReillyBrogan Probably best if you're the one reviewing this.

K1ngfish3r:
`vapoursynth` binary doesn't get installed when using `meson` directly, so it has been switched over to use `meson-python` (again).

mpv doesn't actually pick up vapoursynth @ R74+ so PR brought over as patch until next release comes along

K1ngfish3r:
removed static libs from zimg
moved headers to be listed by the devel package and ensure pkgconfig is provided by only it
add vapoursynth patches from upstream (to be removed in next update)

K1ngfish3r:
updated from R76 -> R77
rebuilt (and tested) ffmpeg and mpv
exported VSSCRIPT_PATH so that mpv can find it properly without the user having to run `vapoursynth register-install`

z/zimg/stone.yaml
```diff
@@ -0,0 +1,23 @@
...
+summary     : Scaling, colorspace conversion, and dithering library
+description : |
+    Project zimg is a light image storage and processing system.
+license     :
+    - WTFPL
+builddeps   :
+    - binary(bison)
+    - binary(flex)
+setup       : |
+    %autogen --disable-static
```
ReillyBrogan: You can use `%reconfigure` instead of `%autogen` as it does slightly less work than the latter. Also move `--disable-static` to a new line.

v/vapoursynth/stone.yaml
```diff
@@ -0,0 +1,63 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : vapoursynth
```
ReillyBrogan, line 4: This looks like it could be split up similar to the Fedora packaging (https://src.fedoraproject.org/rpms/vapoursynth/blob/rawhide/f/vapoursynth.spec)

m/mpv/stone.yaml
```diff
@@ -66,18 +66,31 @@ builddeps   :
...
         -Dlibmpv=true \
         -Dopenal=enabled
 build       : |
     %meson_build
 install     : |
     %meson_install
+
+    # from Arch
+    %install_file TOOLS/{umpv,mpv_identify.sh,stats-conv.py,idet.sh,lua/*} \
+        -D -t %(installroot)%(datadir)/%(name)/scripts
```
ReillyBrogan: That `-D` is already part of the `%install_file` macro

ReillyBrogan (changes requested):

z/zimg/stone.yaml
K1ngfish3r: done

v/vapoursynth/stone.yaml
K1ngfish3r: added python-%(name) and %(name)-tools

m/mpv/stone.yaml
K1ngfish3r: removed

ReillyBrogan (approved):

PR #1703, opened 2026-05-08
swaylock: install /etc/pam.d/swaylock instead of ignoring it

lumi-me-not:
**Summary**

This fixes PAM configuration for swaylock, making it possible to unlock the session.

**Test Plan**

Package was built and installed. Unlocking works correctly.

**Checklist**


ermo:
Technically speaking, this should have a stateless patch that lets %(vendordir) be overridden by %(sysconfdir) configuration files.

lumi-me-not:
I've modified the meson configuration to output pam configs to `/usr/share/defaults/pam.d` and I've modified the swaylock code to check `/usr/share/defaults/swaylock/config` after checking everything else.

lumi-me-not:
I could also pass `PREFIX` into the C code from the meson configuration and use `PREFIX "/share/defaults/swaylock/config",`, if that's desired.

ermo (approved):
LGTM, thanks!

PR #1704, opened 2026-05-09
ferrite: Add at v0.1.0

christian-bendiksen:
**Summary**

Declarative linking tool for the Oxidize ecosystem.
Used for symlinking files through a `.toml` file.
No longer need to wire up things through confusing bash scripts.
Can be used by others who would like to customize with oxidize theming tools as well.

**Test Plan**

Tested on hardware

**Checklist**


ermo (approved):
Interesting!

PR #1709, opened 2026-05-10
opengamepadui-installer: Update Distro Logo

hphilm:
**Summary**

- Update AerynOS branding
- Lock godot to v4.4.1

**Test Plan**

- tested on OrangePi Neo

**Checklist**


ermo (approved):
Too bad we need to vendor godot, but hey ho.

PR #1721, opened 2026-05-14
nushell: Update to 112.2

Samonitari:
**Summary**

nushell Update (from 0.109.1) to 0.112.2

**Test Plan**

build, and run as a shell

**Checklist**

**Discussion**

Letting nushell being stale is a bit problematic.
Although nothing ground-breaking is happening, as it is not stable, stuff oftentimes get deprecated, generating a warning. Jumping from .109 to .112 potentionally lead to breaking changes (although nushell releases have breaking changesanyway) that could have been reacted to in a more timely manner.
I'll keep doing updating the package after some testing quite happily BTW


ermo (approved):
LGTM, thanks!

PR #1729, opened 2026-05-16
oxidize: Update to v0.3.2

christian-bendiksen:
**Summary**

- Introduced a GIO module that gets injected into every GLib-based applications via GIO_EXTRA_MODULES, giving Oxidize live GTK CSS reloading capabilities without any per-app changes.
- Added a workaround fix for Chromium browsers. Now chromium browsers will also hot reload to the selected theme.

**Test Plan**

Tested on hardware

**Checklist**


o/oxidize/stone.yaml
```diff
@@ -2,21 +2,29 @@
...
 license     : MIT
 networking  : true
+builddeps   :
+    - pkgconfig(gio-2.0)
+    - pkgconfig(glib-2.0)
 setup       : |
     %cargo_fetch
 build       : |
     %cargo_build
+    cc $(pkg-config --cflags glib-2.0 gio-2.0) \
```
ermo, line 22: Is there any chance you could be convinced to use our CFLAGS here?

christian-bendiksen: Yes, have added the CFLAGS and confirmed working!

ermo (approved):

PR #1733, opened 2026-05-19
add: ananicy-cpp at v1.2.0 & cachyos-ananicy-rules at v1.1.37

witchlliee:
**Summary**

Add a auto nice daemon for better performance.

NOTE: ananicy-cpp shouldn't be used together with gamemode.

**Test Plan**
Install and enable the service with `sudo systemctl enable --now ananicy-cpp`

**Checklist**


witchlliee:
this should be good to go!

~~scratch that. it seems it doesn't recognize `/usr/share/defaults/ananicy.d` as a path yet~~

K1ngfish3r:
You would need to separate this into two commits, one for each package

PR #1735, opened 2026-05-19
audit: Add at v4.1.4

HarshRelapse:
**Summary**

Audit audits via auditd.

**Test Plan**

This is the test plan that I used to verify that audit works.

```bash
sudo systemctl start auditd.service

sudo systemctl status auditd.service
# Check for errors such as not finding /etc/audit/audit.conf or space is full
```

```bash
# Add watch rule
sudo auditctl -a always,exit -F path=/etc/hostname -F perm=w -F key=aerynos_packaging_test

# Trigger a event
sudo touch /etc/hostname

# Query the logs for the change
# In the output, inside type=SYSCALL there is exe="/usr/bin/touch" as example
sudo ausearch -k aerynos_packaging_test

# Remove the rule for cleanup
sudo auditctl -d always,exit -F path=/etc/hostname -F perm=w -F key=aerynos_packaging_test
```

**Checklist**


ermo:
If you're serious about this whole symlinking business (which looks fairly insane), you probably should not configure things to use `sysconfdir=%(vendordir)`, but rather set `sysconfdir=%(sysconfdir)`.

It would be much nicer if audit could be made actually stateless.

HarshRelapse:
The whole symlinking workaround is unfortunately required since I couldn't find a stateless patch. Fedora and Arch Linux also handle this exactly the same way.

HarshRelapse:
I've decided to remove all plugins and just focus on a minimal audit, it should be on par with solus in terms of how it is packaged. /etc is hardcoded in some places, so changing the sysconfdir only changes where the default config files are placed, not where they are used (stays /etc). The patch could be dropped since no plugins are going to exist, but on the other side, keeping it for the future when someone needs plugins and works on it would make it easier. It's tested based on the test plan.

ermo (commented):
This looks useful. However, for this to land, we need you to commit to maintaining it.

HarshRelapse:
Every PR from me that got merged is on my watch and will be mantained by me. If something is outdated or an issue is there, you can ping me on zulip and It will be my priority to fix any issues.

ReillyBrogan:
FYI the reason this has stalled is that I have been working on stateless support in libaudit. This PR uses the factory pattern which is not actual statelessness. If a future package maintainer updates the config then existing users won't see the changes because the config files in /etc won't be overwritten.

TBH I think it's fairly unlikely that we'll merge this PR as by the time that patch is ready what I have locally will be so different than this that it would be difficult to merge the efforts rather than just using what I have.

PR #1746, opened 2026-05-21
[NFC] discord: update monitoring

K1ngfish3r:
**Summary**

added a link to check for updates and added results from cpesearch

**Test Plan**

url was checked for updates

**Checklist**


ermo (approved):

ReillyBrogan:
That's not an RSS feed though

PR #1747, opened 2026-05-21
gst-thumbnailers: Add at v1.0.0

HarshRelapse:
**Summary**

Packaged gst-thumbnailers, a new GStreamer-based thumbnailing service for Nautilus from Gnome written in Rust

**Test Plan**

Could install but not test, same with ffmpegthumbnailer

**Checklist**


HarshRelapse:
Well, I thought ffmpegthumbnailer was still the default thumbnailer for Gnome and Nautilus. Well, this is the new replacement, which is installed by default in archlinux gnome meta package :/

ermo (approved):

PR #1753, opened 2026-05-23
jack: Add at v1.9.22

jn-sena:
**Summary**

Adds jack2, the audio server. (I needed it as a library dependency for a binary.)

**Test Plan**

Installed and ran the server via `jackd -r -d alsa`. Though, I have never used JACK the server extensively so I didn't do much else with it.

**Checklist**


ReillyBrogan (approved):

ReillyBrogan:
I messed with this a bit. The thing with JACK is that Pipewire provides a JACK server that is actually still maintained, so I basically just stripped this down to the JACK client library which is basically how Pulseaudio is built as well.

ReillyBrogan:
If need be we can add the rest in the future

PR #1759, opened 2026-05-24
dolphin-plugins: Add at v26.04.1

HarshRelapse:
**Summary**

Added dolphin-plugins.

**Test Plan**

Tested by turning on the git plugin in dolphin and looking for changes to git files.

**Checklist**


ermo (approved):
Looks very useful.

PR #1765, opened 2026-05-25
jay: add at v1.13.0

witchlliee:
**Summary**

Yet another wayland compositor (Window Manager) built on rust. This one is way ahead in features tho (It implements most of the wayland protocols, also has Vulkan rendering and HDR etc).  

Note: I think I'm submitting too much packages already xD. I will stop for a bit. I swear.

**Test Plan**

Install, run `jay config init` for it to write the default config in `~/.config/jay/config.toml`, then you can make your changes or you could use the jay's control center to configure it via a GUI (The changes aren't permanent tho. Only for the current session). You can bring up the control center with `Alt+C` while logged in jay.

**Checklist**


mauvray:
@witchlliee Hey, I just saw your MR (maybe too late ^^'). How did you find out the dependencies of jay ? I had a stone file for myself, and I did not have exactly the same dependencies, although it was for jay 1.12.0.

ermo:
We could probably (and honestly) benefit from jay going in if someone else wants to take this PR over?

mauvray:
I can submit my own recipe, I guess I should need to check it again, to make sure I do not bring unneeded dependencies. I can't really provide a time estimation more precise than next week or the week after though, if it's okay for you @ermo

PR #1768, opened 2026-05-25
otter-launcher: Add at v0.7.4

TheXnlyGhxst:
**Summary**

* Add otter-launcher, a hackable cli/tui launcher built for keyboard-centric WMs. Made with rust.

**Test Plan**

* Build succeeds -> install succeeds -> all functionality appears to work

**Checklist**


ermo (approved):
Sorry, it looks like we completely missed this. Nice find!

PR #1770, opened 2026-05-26
immer: Add at v0.9.1

witchlliee:
**Summary**

Add immer. I'm working on packaging krita (which i think we really should have it) but it depends on immer as dependency.

**Test Plan**

**Checklist**


HarshRelapse:
There is a small problem in all PRs: Zed by default formats YAML code, but it should not get formatted, so it looks completely different from what it should look like ([nano as an example](https://aerynos.dev/packaging/workflow/creating-a-new-recipe/)). So please disable Zed's default format on save and reformat all PRs to the standard style.

First part for what naming to use for commit and title https://aerynos.dev/packaging/workflow/submitting-a-pr/

witchlliee:
> There is a small problem in all PRs: Zed by default formats YAML code, but it should not get formatted, so it looks completely different from what it should look like ([nano as an example](https://aerynos.dev/packaging/workflow/creating-a-new-recipe/)). So please disable Zed's default format on save and reformat all PRs to the standard style.
> 
> First part for what naming to use for commit and title https://aerynos.dev/packaging/workflow/submitting-a-pr/

I think i disable the format on save. Can you take a look and tell me if it looks alright now?

witchlliee:
All done then. It should be all good and clean at this point

PR #1782, opened 2026-05-28
protobuf: Update to v35.0

witchlliee:
**Summary**

Update protobuf to the latest release and add new dependency bazelisk.

**Test Plan**

- Confirm its functionality 

**Checklist**


K1ngfish3r:
I recommend adding the rebuilds to this as well since some packages depend on protobuf (if not, might cause version mismatch errors like it happened for me on opencv) 

Also, technically Bazelisk should come first in the commit history, not sure if it is relevant

p/protobuf/stone.yaml
```diff
@@ -2,19 +2,25 @@
...
 builddeps   :
+    - binary(bazel)
     - cmake(absl)
+    - pkgconfig(python3)
     - pkgconfig(zlib)
+    - python(build)
+    - python(wheel)
+networking  : true
 setup       : |
+    bazel --version
```
K1ngfish3r, line 23:
```suggestion
    bazel fetch //python/dist:binary_wheel
```
Make it more similar to cargo macros

p/protobuf/stone.yaml
```diff
@@ -24,8 +30,11 @@ setup       : |
         -Dprotobuf_LOCAL_DEPENDENCIES_ONLY=OFF
 build       : |
     %cmake_build
+    bazel build //python/dist:binary_wheel
```
K1ngfish3r, line 33:
```suggestion
    bazel build --nofetch //python/dist:binary_wheel
```
Since we fetch earlier

K1ngfish3r (commented):
Make more similar to cargo macros

witchlliee:
> I recommend adding the rebuilds to this as well since some packages depend on protobuf (if not, might cause version mismatch errors like it happened for me on opencv)
> 
> Also, technically Bazelisk should come first in the commit history, not sure if it is relevant

Wdym by the rebuilds?

ermo:
> > I recommend adding the rebuilds to this as well since some packages depend on protobuf (if not, might cause version mismatch errors like it happened for me on opencv)
> > Also, technically Bazelisk should come first in the commit history, not sure if it is relevant
> 
> Wdym by the rebuilds?

@witchlliee There are packages that depend on the package you are building, in the sense that they link against the ELF symbols exposed by it. This concept is also known as "reverse dependencies".

If you update the protobuf package, you also need to update its reverse dependencies where applicable. If you can't figure out what the reverse dependencies are for protobuf, you likely should not be attempting to update it in the first place, and should leave it to more experienced packagers.

FWIW, this concept repeats all across the repository, and packagers are expected to understand said concept before proposing updates to packages that have reverse dependencies.

witchlliee:
> > > I recommend adding the rebuilds to this as well since some packages depend on protobuf (if not, might cause version mismatch errors like it happened for me on opencv)
> > > Also, technically Bazelisk should come first in the commit history, not sure if it is relevant
> > 
> > 
> > Wdym by the rebuilds?
> 
> @witchlliee There are packages that depend on the package you are building, in the sense that they link against the ELF symbols exposed by it. This concept is also known as "reverse dependencies".
> 
> If you update the protobuf package, you also need to update its reverse dependencies where applicable. If you can't figure out what the reverse dependencies are for protobuf, you likely should not be attempting to update it in the first place, and should leave it to more experienced packagers.
> 
> FWIW, this concept repeats all across the repository, and packagers are expected to understand said concept before proposing updates to packages that have reverse dependencies.

Ah, i understand what you mean now. It seems simple enough but only if i could tell every package that depends on it, which could be too much, i suppose. I will let this one for someone more experienced then

PR #1783, opened 2026-05-28
ffmpegthumbnailer: Deprecate (use gst-thumbnailers instead)

HarshRelapse:
**Summary**

I thought ffmpegthumbnailer was still the default thumbnailer for Gnome and Nautilus. Well, this is the new replacement, which is installed by default in archlinux gnome meta package :/

gst-thumbnailers was added as rundep for nautilus so removing ffmpegthumbnailer should cause no issue.

**Test Plan**

**Checklist**


hphilm:
Any removal needs to be documented: https://github.com/AerynOS/recipes/blob/main/removals.md and checked if any remaining packages may still need it.

HarshRelapse:
There aren't any deps based on ffmpegthumbnailer or requiring it. There is no problem in removing it. Updated the removals.md.

ermo:
@HarshRelapse : 

You need to make ffmpegthumbnailer be a package that contains only a README file + rundeps on gst-thumbnailers (so people who installed it will get auto-upgraded to gst-thumbnailers instead).

Look at other deprecated packages for how we currently handle this.

HarshRelapse:
Okay, updated to specification

ermo (approved):
There we go.

PR #1785, opened 2026-05-28
niri: Fix xwl-s dep should be optional

witchlliee:
**Summary**

It takes out xwl-s from dependencies since i really think this one should be optional specially because it's not built-in xwayland like in other desktop environments and also nowdays most of the things runs on native wayland anyway. It can also be frustrating if someone just wants to run a pure wayland-session.

**Test Plan**

- See how it doesn't pull xwayland-satellite

**Resolves**:
(If applicable for the recipes repository) Resolves aerynos/recipes

**Checklist**


ReillyBrogan:
IMO users installing Niri are going to expect xwayland applications (like Steam) to work out of the box. At some point we'll support a `recommends` key for soft dependencies but for right now that's not possible so I would prefer to leave this.

PR #1787, opened 2026-05-29
python: Don't nuke ensurepip's bundled pip wheel

joebonrichie:
**Summary**

Fixes `python -m ensurepip` which is used by venv environments

**Test Plan**

Setting up venv scripts on an existing project now works as expected

**Checklist**


ReillyBrogan:
I don't think this is the right solution. We ship wheels for pip in python-pip-wheel which is a rundep of python-virtualenv. I suspect we just need a patch for python-virtualenv or something.

ReillyBrogan:
Actual issue fixed by https://github.com/AerynOS/recipes/commit/40ec97c9fcbeba407c6a458ba24550ca2e528b27

PR #1796, opened 2026-05-31
keepsecret: Add at v1.1.0

HarshRelapse:
**Summary**

Added KeepSecret as a possible KWalletManager replacement, since KDE Linux replaced KWallet with KeepSecret.
https://invent.kde.org/kde-linux/kde-linux/-/merge_requests/504

An important note: If you change the plasma pkgset to use KeepSecret instead of kWalletManager, you must also explicitly add kf6-kwallet (or oo7 one day) next to KeepSecret, else the plasma pkgset will pull in KeepSecret without kf6-kwallet.

**Test Plan**

Installed, checked to see if the store could be locked and unlocked, checked to see if the password could be copied, created a test password, and deleted it.

**Checklist**


hphilm:
Is `kf6-kwallet` a hard dependency to make it run or an optional dependency but needed to make Plasma work as usual?

HarshRelapse:
Since KeepSecret also supports keepassxc and oo7, it is agnostic and does not have a hard dependency on kf6-kwallet (it's optional). KeepSecret doesn't need kf6-kwallet, but plasma does.

PR #1815, opened 2026-06-10
greetd: Fix not being able to use other greeters

witchlliee:
**Summary**

It fix the lacking of a `.service` and also add a stateless patch to use the config.toml distributed in `/usr/share/defaults/greetd/config.toml` in case `/etc/greetd/config.toml` doesn't exists.

**Test Plan**

- Enable the service with `sudo systemctl enable greetd` (disable the one you were using first ofc)
- Install and configure a greeter like [tuigreet](https://github.com/apognu/tuigreet)

**Checklist**


hphilm:
Provided patch doesn't do anything. Check again and provide the proper one. Also CI claims to reorder builddeps.

witchlliee:
> Provided patch doesn't do anything. Check again and provide the proper one. Also CI claims to reorder builddeps.

Tbh I think maybe the patch isn't even needed since the user needs to create `/etc/greetd/config.toml` either way to setup their greeter. So what is the point?

K1ngfish3r:
ie: patch file is empty

witchlliee:
> ie: patch file is empty

Oh. Sorry, it should be fixed now. I hate when this hapens xD

witchlliee:
It seems sometimes when I restart my system, greetd.service isn't enabled anymore so it throws me to tty and I have to enable greetd again. This is VERY frustrating and idk why this happens.

ReillyBrogan:
dd4bcbd7669fbe149411907cc1c3b5404d3265ec

PR #1820, opened 2026-06-10
uwsm: Add at v0.26.4

witchlliee:
**Summary**

Add uwsm. Useful for Wayland Compositors that doesn't integrate systemd natively. Also Hyprland provides a (uwsm-managed) session ootb.

**Test Plan**

- Log into Hyprland (uwsm-managed)

**Checklist**


u/uwsm/stone.yaml
```diff
@@ -0,0 +1,31 @@
...
+release     : 1
+homepage    : https://github.com/Vladimir-csp/uwsm
+upstreams   :
+    - https://github.com/Vladimir-csp/uwsm/archive/refs/tags/v0.26.4.tar.gz : b7eeb03ab3950db933d3561d229ee9da55235f7f9ff5e481d6e90b3288f52954
+summary     : Universal Wayland Session Manager
+description : |
+    Provides a set of Systemd units and helpers to set up the environment and manage standalone Wayland compositor sessions.
+license     :
+    - MIT
+rundeps     :
```
Jaredy899, line 15: Rundeps need to be addressed for CI please.

u/uwsm/stone.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - python-pyxdg
+    - python-dbus
+builddeps   :
```
Jaredy899, line 18: Same here!

Jaredy899 (changes requested):
I'm unclear what this patch does?

Jaredy899:
@witchlliee Were you planning on landing this? If not, we should close until we're ready.

witchlliee:
> @witchlliee Were you planning on landing this? If not, we should close until we're ready.

I was at some point but pyxdg is incompatible with newer python versions, therefore, the patch. Not sure if it's really worth it. I can rework this if you plan in adding it tho.

Jaredy899:
The patch is empty. It doesn't do anything.

witchlliee:
> The patch is empty. It doesn't do anything.

Oh god. That was a long time ago. Was still learning, sorry

PR #1825, opened 2026-06-12
amneziawg-module: add at v3.1.20260812

jn-sena:
**Summary**

Adds the AmneziaWG kernel module!

This is my first kernel module, so I'm sorry if I've done anything stupid-. I've tried to make it similar to `zfs-module` and `nvidia-open-gpu-kernel-modules`.

For 7.1.5+, I had to include the patch from https://github.com/amnezia-vpn/amneziawg-linux-kernel-module/pull/194.

Also had to manually add the sublevel to the make macro, because it wouldn't build otherwise for plain 7.1. (Not needed as of 7.1.1 and newer, but left it in to be safe in the future.)

**Test Plan**

I've installed the kernel module for my current kernel (`gaming`), rebooted, and connected to my peer using it and `amneziawg-tools`.

**Checklist**


ermo (commented):
For this to go in, we need you to commit to maintaining it.

jn-sena:
Well of course, I'd be happy to. Do I need to do something specific?

ermo:
That fact that you just committed to help maintaining it is enough.

@ReillyBrogan Any objections?

PR #1826, opened 2026-06-12
amneziawg-tools: add at v3.1.20260812

jn-sena:
**Summary**

Adds the AmneziaWG counterpart of `wireguard-tools`.

**Test Plan**

I've connected to my peer using `awg-quick` app and it's systemd service template. It worked with both `amneziawg-go` and the kernel module.

**Checklist**


ermo (commented):
Fort this to land, we need you to commit to updating it.

jn-sena:
I do commit to it, alongside amneziawg-module.

PR #1839, opened 2026-06-17
kwin: fix graphical glitches

hphilm:
**Summary**

image

Some newer kernels might create graphical glitches in Plasma 6.7. Adding this workaround helps due to kernel bugs.
Info: https://mail.kde.org/pipermail/distributions/2026-June/001711.html

Applying the kernel patch seems not to fix it. The workaround does.

**Test Plan**

- install and check for regressions

**Checklist**


hphilm:
Repo for testing: `sudo moss repo add plasma https://cdn.aerynos.org/repo/plasma/x86_64/stone.index -p 20`

ReillyBrogan (approved):

HarshRelapse:
Worked with my system :)

PR #1857, opened 2026-06-19
vesktop: Add at v1.6.5

witchlliee:
**Summary**

Custom discord client with better linux support.

**Test Plan**

- Launch it

**Checklist**


NomadicCore:
@ReillyBrogan @ermo 

There was a bit of a discussion around Vesktop in Zulip.

One of the things I noted is that this recipe currently renames the rpm rather than building from source.

I would guess that we would have a rule in place to build from source first as it ensures ownership of outputted packages coming from our build infrastructure but wanted to check my logic here?

As a separate point, are we open to including Vesktop when we have Discord in the repo? I use the Vesktop flatpak image and having it natively in our repo would be cool to get away a little from discord enshitification?

ermo:
As long as the from-source rebuild dep chain is not insane, this is certainly something we could entertain.

Jaredy899:
Superseded by #2031

PR #1866, opened 2026-06-20
README.md: clarify "metrics-based performance improvements" quip

upslaught:
**Summary**
this clarifies that the performance of **packages** would be improved based on the metrics, I was confused by what the performance of would be improved upon (moss? the system?)


upslaught:
(I'm actually unsure whether this is the right way to title this pr)

ermo (approved):
LGTM, thanks!

PR #1872, opened 2026-06-22
lzip: Rebuild to test vessel v0 repo format upgrade

tarkah:
**Summary**

Bump to check if we can build on the upgraded repo and everything gets updated correctly.

**Test Plan**

Check that everything looks sane after we land this.

**Checklist**


ermo (approved):
Let's see.

PR #1874, opened 2026-06-23
steam-devices: revert a regression with inputplumber

hphilm:
**Summary**

The PR https://github.com/ValveSoftware/steam-devices/pull/83 introduced a regression with `inputplumber` and its `SteamOS Handheld Controller` profile. Revert this for now till upstream provides a solution

**Test Plan**

- tested on OrangePi Neo device

**Checklist**


hphilm:
`deck` seems to work, similar as `xbox-elite` with the new rules. only `deck-uhid` doesn't. Tested via `inputplumber device 0 targets set <deck,deck-uhid,xbox-elite>`. Derek informed Valve about it ...

ReillyBrogan (approved):

hphilm:
See also: https://github.com/ValveSoftware/steam-devices/issues/86

PR #1888, opened 2026-06-26
ocl-icd: export vendor dir for stateless

K1ngfish3r:
**Summary**

need to `export OCL_ICD_VENDORS=/usr/share/OpenCL/vendors` for a `venv` so that OpenCL devices can be found
taken from l/less

**Test Plan**

run pytorch app

**Checklist**


ermo (approved):
Fingers crossed?

PR #1901, opened 2026-06-29
iio-niri: add at v2.3.1

jn-sena:
**Summary**

Add [iio-niri](https://github.com/Zhaith-Izaliel/iio-niri) for automatic screen orientation with supporting devices in Niri!

Patch from https://github.com/Zhaith-Izaliel/iio-niri/pull/13 for the `--locked` flag.

**Test Plan**

Install, run and rotate around while running Niri.

**Checklist**


i/iio-niri/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id: ~ # Last checked (2026-06-29)
```
ermo: Please create an id for this package over at https://release-monitoring.org.

You can log in with your GH account.

ermo (commented):
Almost ready, except it needs a monitoring id.

ermo (commented):
For this to land, we need you to commit to updating it.

jn-sena:
> For this to land, we need you to commit to updating it.

I do commit to it

Jaredy899:
Is there a check that can be added to this? (Note that some cargo tests fail and aren't worth adding but if it does, we should). Otherwise seems simple enough.

ermo (approved):
Looks like a nice addition with not too many revdeps.

PR #1904, opened 2026-06-30
Plasma: Update to v6.7.2

hphilm:
**Summary**

Announcement: https://kde.org/announcements/plasma/6/6.7.2/
Info: https://kde.org/info/plasma-6.7.2/

**Test Plan**

**Checklist**


ermo (commented):
Are the CI builddep ordering complaints valid?

hphilm:
Lets see if current changes may satisfy the CI checks ...

ermo:
Reilly used to split these up into two parts, mainly because the summit resolver haz the dumb. He'd then land the first part and see that build, and then land the 2nd part such that the packages in that would correctly be built against the first part.

ermo (approved):
Let's give it a go.

PR #1931, opened 2026-07-04
moss: Ignore notices in command-not-found handler

sugar700:
Without this change, typing an unknown command in shell that available in repositories results in `Install package '"/etc/moss/system-model.kdl"'?` when `system-model.kdl` is in `/etc/moss`.

**Summary**

This pull request changes `2>&1` to instead completely ignore errors.

**Test Plan**

I sourced the command not found handler, and ran a command that wasn't installed on my computer (`node`), both with `/etc/moss/system-model.kdl` file, and without it.

**Checklist**


ermo (approved):
Nice catch. Cheers!

PR #1937, opened 2026-07-04
dust: Add at 1.2.4

### LINKED ISSUE #1913: [Package Request]: dust
### Existing package request?

### Homepage of package

https://github.com/bootandy/dust

du + rust = dust. Like du but more intuitive.

### Licence of package?

[Apache-2.0 license](https://github.com/bootandy/dust?tab=Apache-2.0-1-ov-file#)

https://github.com/bootandy/dust/archive/refs/tags/v1.2.4.tar.gz

K1ngfish3r:
**Summary**

Fixes https://github.com/AerynOS/recipes/issues/1913

From README
>du + rust = dust. Like du but more intuitive.
Because I want an easy way to see where my disk is being used.
Example
> * We see `target` has 1.8G
> * `target/debug` is the same size as `target` - so we know nearly all the disk usage of the 1.8G is in this folder
> * `target/debug/deps` this is 1.2G - Note the bar jumps down to 70% to indicate that most disk usage is here but not all.
> * `target/debug/deps/dust-e78c9f87a17f24f3` - This is the largest file in this folder, but it is only 46M - Note the bar jumps down to 3% to indicate the file is small.
> 
> From here we can conclude:
>  * `target/debug/deps` takes the majority of the space in `target` and that `target/debug/deps` has a large number of relatively small files.

**Test Plan**

Requesting @Jaredy899 for test. am not familiar with package

**Checklist**


Jaredy899:
Rest of the package works fine. Thanks.

Jaredy899:
Actually there are man pages that need to be added.

K1ngfish3r:
updated with manpage

Jaredy899:
Will you also fix the Summary so it says what `dust` does.
and make so it says Resolves or fixes #1913

K1ngfish3r:
updated

Jaredy899:
Can you use the macro for the elvish completion?

K1ngfish3r:
i believe we dont have an elvish terminal (???)
fixed the macro tho

Jaredy899:
Thank you! Since elivish can be added manually and we have the macro, might as well use it.

ermo (approved):
LGTM, thanks!

PR #1940, opened 2026-07-05
justfile: Replace deprecated env_var_or_default with env

sugar700:
According to https://just.systems/man/en/functions.html#environment-variables, `env_var_or_default` function is deprecated.

**Summary**

Uses `env`, which isn't deprecated.

This was pointed by my text editor, so I figured, may as well do this change to make it not point this.

**Test Plan**

Tried to use `just` after this change.

**Checklist**


ermo (approved):
Nice catch!

PR #1947, opened 2026-07-05
[RFC] linux-stable: Allow out of tree modules to survive

joebonrichie:
**Summary**

Allow linux module updates to survive moss state changes using the same trick as `/usr/local`

**Test Plan**

out of tree installed module `framework-laptop-kmod` now survives moss state changes

**Checklist**


hphilm:
So this would partly enable dkms support as well? Since we already have some extramodules like Nvidia driver compiled per kernel update, do we see any issues adding this as well?

joebonrichie:
It look's like our default `make modules_install` behavior installs to `/lib/modules/<kernel-version>/updates` but we install packaged out of tree modules e.g nvidia to `/lib/modules/<kernel-version>/extra`

Currently the user will still have to manually rebuild and install the module if the kernel is updated but could take a look at getting dkms working as well. All this change does for the moment is add an escape hatch for the `updates` module folder so it doesn't get wiped with each moss transaction.

ReillyBrogan:
Wouldn't it make more sense to use [depmod.d](https://man7.org/linux/man-pages/man5/depmod.d.5.html) configuration instead?

joebonrichie:
Yeah that's definitely an option, could add a default depmod conf e.g.

```
/etc/depmod.d/usermods.conf
external * /var/lib/kernel-modules
```

The user can override `INSTALL_MOD_PATH` to `make modules_install` to install to the var location and that should work

I think the tricky bit is documentation/getting this to work OOTB, you'd probably want `make modules_install` to just work instead of being a special snowflake distribution.

hphilm:
If with this current change enables something without a fuzz why not merge it as is ...

joebonrichie:
There's an issue where stale directory entries remain in /var once the kernel has been updated, hold off for now.

ReillyBrogan:
> I think the tricky bit is documentation/getting this to work OOTB

My strong preference is adding documentation for this instead.

PR #1983, opened 2026-07-13
podman stack update

### LINKED ISSUE #1811: Podman package problems: quadlet doesn't search user folders
I haven't done much search as of now, but services files don't get generated from container files by quadlet.
quadlet should find container files in `~/.config/containers/systemd/`

But 
``` shell
> /usr/lib/podman/quadlet -dryrun user
quadlet-generator[33271]: No files parsed from [/run/containers/systemd /etc/containers/systemd /usr/share/containers/systemd]
```

one thing I observed that on OpenSUSE Tumbleweed, `quadlet` `rootlessport` (as well as `catatonit`) get installed to `/usr/libexec/podman` - as well as `netavark` and `aardvark-dns`

### LINKED ISSUE #1860: podman doesn't read configs from /usr/share/defaults/containers/
I'm on unstable, and `podman pull` fails like this:
```
$ podman pull archlinux:base-devel
Error: short-name "archlinux:base-devel" did not resolve to an alias and no containers-registries.conf(5) was found
$ podman pull docker.io/archlinux:base-devel
Error: no policy.json file found at any of the following: "/home/yalter/.config/containers/policy.json", "/etc/containers/policy.json"
```
The file exists in `containers-common` which is installed, however it's in `/usr/share/defaults/containers/`.

If I do `sudo ln -s /usr/share/defaults/containers /etc/containers` then it starts working fine.

The short-name issue is probably separate and needs some common registry defined like docker.io.

Jaredy899:
**Summary**
- Release notes can be found [here](https://github.com/podman-container-tools/podman/releases#release-v6.0.1).
- Resolves: #1860 
- Resolves: #1811 

Signed-off-by: Jared Cervantes <jared@jaredcervantes.com>

**Test Plan**
`podman pull alpine:latest`
`podman run --rm alpine:latest uname -a`

**Checklist**


ermo (approved):
\o/

PR #1990, opened 2026-07-14
gnist: Add at v0.3.5

christian-bendiksen:
**Summary**

### Bug fixes and rebrand of Oxidize
`gdk_display_manager_get()` was a fatal error when called
before `gtk_init()`. We hit this whenever it loaded into a GTK4
process that never initialized GTK.

- resolve `gtk_is_initialized` and added `gtk_initialized()` helper
  that reports true on GTK3
- `attach_or_retry()` now bails into the existing 5x100ms retry
  loop while GTK is unitialized
- the `display-opened` wait is now GTK3-only. On GTK4 a missing
  display post-init can't happen, so it just logs and skips

**Test Plan**

Tested on hardware, fixed.

**Checklist**


ermo (commented):
@christian-bendiksen :

Would you be open to adding the removals/renames to this PR in separate commits, so it's obvious what is going on for future reference?

ermo (dismissed):

ermo:
@christian-bendiksen Turns out your original approach was the correct one (doing removals in one go). Could you rebase this please?

ermo (approved):

PR #1989, opened 2026-07-14
malm: Add at v0.3.2

christian-bendiksen:
**Summary**

### Huge update and rebrand of Ferrite
- Declarative KDL configuration using reusable modules, typed inputs,
  requirements, and outputs.
- Typed rendering for JSON,  JSONC, TOML, INI, text, Lua, KDL, XML, and CSS
- Output validation for formats including shell, Lua, CSS, GTK CSS, KDL, TOML,
  JSON, Hyprland, Mango, and Kanshi
- Composable profiles with inheritance, abstract profiles, multiple module
  instances, and aliases.
- Provider slots for replacing roles such as compositors without duplicating
  profiles
- Fine-grained patching of records and keyed collections using `set`, `unset`,
  `append`, `replace` and `remove`
- Replacable fragments for profile-specific native configuration files
- Conditional and repeated output through `@when`, `@each`, `@range`, `@spread`,
  and optional values
- File inclusions and interpolation  with `@file`, typed references, formatted
  strings, and built-in variables
- Safe deployment of generated files, repository files, directories, and 
  managed symlinks with conflict backups
- Preview workflow through `malm check`, `render`, `plan` and `apply`
- Independent named states, allowing several configurations to coexist
- Drift detection with `malm status`
- Transactional history and rollback using state logs and `state checkout`
- Crash recovery with journaled operations, `state fsck`, and `state recover`
- State lifecycle management with disable, enable, destroy, pin, unpin,
  prune, and disk-usage reporting
- Remote git configurations pinned by commit, tag, or branch, with tracking
  and updates
- Remote safety controls for local includes, external symlinks, secrets,
  unverified assets, and SSRF-sensitive downloads.
- Managed assets with HTTPS enforcement, SHA-256 verification, archive
  extraction, and content-addressed storage
- Module requirement check through `malm doctor`
- Configuration extensions using `extend-module`, `extend-profile`, globals,
  and optional local includes
- JSON command output for automation and tooling integration

**Test Plan**

Tested on hardware

**Checklist**

- [x] Recipe was built and tested against the volatile stream
- [x] This change could gainfully be highlighted in the Stream Update notes once merged

PR #1991, opened 2026-07-14
pkgset-smia-base: Add at v2026.07.14

christian-bendiksen:
**Summary**

The base pkgset for Smia (The Forge) which is a ready-made
configuration made with Malm and Gnist.

**Test Plan**

Tested on hardware

**Checklist**


NomadicCore:
Hey,

I think we can call the version 2026.07, it doesn't necessarily need to line up with the current iso version (I'm assuming that's why you have used 2026.05.2?)

We can update pkgset versions to the monthly they were created/updated and have there be independent of iso versions.

christian-bendiksen:
> Hey,
> 
> I think we can call the version 2026.07, it doesn't necessarily need to line up with the current iso version (I'm assuming that's why you have used 2026.05.2?)
> 
> We can update pkgset versions to the monthly they were created/updated and have there be independent of iso versions.

I see! Have amended the changes

ermo:
- depends on #1989 
- depends on #1990 

Must not be merged before the above have been merged and finished building.

PR #1999, opened 2026-07-15
cardwire: Add at v0.10.3

hphilm:
**Summary**

Introduction: https://opengamingcollective.github.io/cardwire/
Changelog: https://github.com/OpenGamingCollective/cardwire/releases/tag/v0.10.3
Readme: https://github.com/OpenGamingCollective/cardwire/blob/main/README.md

**Test Plan**

**Checklist**


ermo (approved):
This looks interesting!

For Reilly's benefit:

This is in my view generic enough that it should be allowed to have a non-central systemd-preset that turns it on if installed. It can still be overridden by individual spins (and users), but the benefits from having it turned on ootb when installed are looking pretty undeniable from my PoV.

PR #2007, opened 2026-07-16
forgejo-cli: Add at v0.5.0

BarryLhm:
**Summary**

Adds forgejo-cli, a code forge client like `github-cli`(gh)

**Test Plan**

At least basic functionalities are tested to work fine

**Checklist**

- [X] This change could gainfully be highlighted in the Stream Update notes once merged


BarryLhm:
ooh, it seems code owner confirmation is granted per branch

NomadicCore:
From what I can tell, this will allow us to more easily interact with Codeberg from command line, which is probably really helpful for us!

BarryLhm:
interestingly, my forked repo on gh is actually a push-mirror target of a repo named the same on codeberg.
so to some extent it's one of the first aerynos (related) repos on codeberg :}

f/forgejo-cli/monitoring.yaml
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+releases:
+  id: 377539
+  rss: ~
```
Jaredy899: Hello! Can we change this to 
```yaml
rss: https://codeberg.org/forgejo-contrib/forgejo-cli/releases.rss
```

Jaredy899 (changes requested):

Jaredy899:
Can you squish your commits and add a commit message like?
`fj, a CLI client for Forgejo. It doesn't try to replace your usage of git but working alongside it. It handles all the Forgejo-specific things that git doesn't`

BarryLhm:
> Can you squish your commits and add a commit message like?
> `fj, a CLI client for Forgejo. It doesn't try to replace your usage of git but working alongside it. It handles all the Forgejo-specific things that git doesn't`

squashing needs native git so just finished on termux
commit message seems not a necessity and most existing commits only have a title

Jaredy899:
Please add shell completions.

BarryLhm:
i'm not familiar with this, searching....
seems to need manually calling the binary to generate
gonna get it done tomorrow

Jaredy899:
You can use [arch](https://gitlab.archlinux.org/archlinux/packaging/packages/forgejo-cli/-/blob/main/PKGBUILD?ref_type=heads) as a guide.

BarryLhm:
done

f/forgejo-cli/stone.yaml
```diff
@@ -0,0 +1,28 @@
...
+summary     : CLI application for interacting with forgejo
+description : |
+    fj, a CLI client for Forgejo. It doesn't try to replace your usage of git but working alongside it. It handles all the Forgejo-specific things that git doesn't
+license     :
+    - Apache-2.0
+    - MIT
+builddeps   :
+    - pkgconfig(openssl)
+networking  : true
+setup       : |
```
Jaredy899, line 21: This still needs to be addressed. And we should do it similiar to arch and link against our own libraries. 
```yaml
builddeps:
  - pkgconfig(libgit2)
  - pkgconfig(libssh2)
  - pkgconfig(openssl)
environment: |
  export LIBGIT2_NO_VENDOR=1
  export LIBSSH2_SYS_USE_PKG_CONFIG=1
  export OPENSSL_NO_VENDOR=true
```
`nushell` does the same thing for reference.

Jaredy899 (changes requested):

f/forgejo-cli/stone.yaml
BarryLhm: again, thanks for instruction

BarryLhm: the binary has not changed, but for future-proof finished

BarryLhm:
done

Jaredy899:
Close in favor of #2159

PR #2008, opened 2026-07-16
iwd: Fix iwctl input handling

irevoire:
**Summary**

In the current version, we're using the `editline` library to build iwctl, which apparently doesn't work. See https://aerynos.zulipchat.com/#narrow/channel/543066-General/topic/How.20does.20it.20compare.20to.20arch.20or.20nix/near/610852879 for more info. This commit gets rid of the `editline` dependency, which means we fall back on the `readline` library instead.

**Test Plan**

I did a `just build`, `just mv-local`, `moss sync -u` and then ran the new binary and it works.

I don't know how to "test against the volatile stream". It's not mentioned on this page: https://aerynos.dev/packaging/workflow/submitting-a-pr/
Am I missing something?

**Checklist**


ermo:
> I don't know how to "test against the volatile stream". It's not mentioned on this page: [aerynos.dev/packaging/workflow/submitting-a-pr](https://aerynos.dev/packaging/workflow/submitting-a-pr/)
> Am I missing something?

https://aerynos.dev/packaging/workflow/basic-workflow/#making-moss-use-the-local-repository

irevoire:
Ok, done!
My issue was not about enabling how to enable the volatile stream, but more than I didn't understand the purpose of enabling this channel.
I thought the process was something like:
- As I maintainer I push my package to volatile stream
- Make sure it works
- It gets pushed to the unstable stream

But actually you just want me to build my package while using all the dependencies from the volatile stream right?
In that case I just did it and it works :+1: 
(and yes, I ran moss sync -u which made a ton of update)

ermo:
> you just want me to build my package while using all the dependencies from the volatile stream right?

Boulder builds against our volatile stream by default.

We want you to ***test*** the packages you build while you have the volatile stream enabled (otherwise, the packages would not be tested in the environment in which they were built).

ermo:
Please fix the CI warning re. sorting of builddeps.

ReillyBrogan (changes requested):
Please leave a comment in the recipe explaining the issue and that using readline instead of libedit fixes it.

irevoire:
Is that enough?

ReillyBrogan:
Perhaps `# Use readline instead of libedit as the latter has terminal rendering issues with iwctl` and please move it to above `%configure`

irevoire:
Hey, I proposed another sentence because;
1. I'm pretty sure the issue comes from the way iwctl uses libedit instead of libedit directly
2. It was not rendering issue but really the way the inputs were sent. If it was only rendering issue I could've used it, but here I _thought_ it was working but my issue was that iwctl was not able to get my password correctly.

Also on another subject, is there a place to make issue on the "kernel" projects? Because the bug is fairly easy to reproduce but I just don't know where you're supposed to post it ahah

EDIT:
Also if you want me to write another sentence or use exactly what you proposed let me know, I'm not too stubborn

ReillyBrogan:
> Also on another subject, is there a place to make issue on the "kernel" projects? Because the bug is fairly easy to reproduce but I just don't know where you're supposed to post it ahah

Well it looks like it was reported a few years ago without much traction on it: https://lore.kernel.org/iwd/?q=libedit

Honestly readline is the more advanced and tested library anyway so I wouldn't worry about it too much. Honestly I'm a bit surprised that libedit is the default if readline is also found as that's a bit baffling.

ReillyBrogan:
Ah I see libedit is actually only the default if you explicitly use `--enable-libedit`

ReillyBrogan (approved):

irevoire:
Yep definitely not the default ahah

PR #2009, opened 2026-07-16
iw: Fix building

irevoire:
**Summary**

iw depends on gzip but it was not included in the dependencies.

**Test Plan**

/!\ The package was **not** properly tested.
I only made sure I could run it, but I don't know how to use it.
I did build it on the volatile channel though

**Checklist**

- [x] Recipe was built ~~and tested~~ against the volatile stream


ReillyBrogan (approved):

ReillyBrogan (changes requested):
You need to rebuild the package so that the manifest has the updated relnum

irevoire:
Ah yes my bad, I did the bump _after_ the build sorry

ReillyBrogan (approved):

PR #2015, opened 2026-07-17
Add aha and update kinfocenter dependency

BarryLhm:
**Summary**

whyever opened kinfocenter and it said missing dependency of aha, so packaged it

**Test Plan**

It's tested against it's original reason for packaging (kinfocenter), working fine.

**Checklist**


a/aha/stone.yaml
```diff
@@ -0,0 +1,20 @@
...
+summary     : ANSI HTML Adapter
+description : |
+    Converts ANSI excape sequences of a unix terminal to HTML code.
+license     :
+    - LGPL-2.0-or-later
+    - MPL-1.1
+build       : |
+    %make
+install     : |
+    %install_bin aha
```
Jaredy899: Here you can do `%make_install PREFIX=%(prefix)` and it will install aha and the man pages so you can delete the manual install of the man page.

a/aha/stone.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+name        : aha
+version     : "0.5.1"
+release     : 1
+homepage    : https://github.com/theZiz/aha
+upstreams   :
+    - https://github.com/theZiz/aha/archive/refs/tags/0.5.1.tar.gz : 6aea13487f6b5c3e453a447a67345f8095282f5acd97344466816b05ebd0b3b1
+summary     : ANSI HTML Adapter
+description : |
+    Converts ANSI excape sequences of a unix terminal to HTML code.
```
Jaredy899: "escape" instead of "excape"

Jaredy899 (changes requested):

Jaredy899 (changes requested):
Since you rebuilt `kinfocenter`, you'll have to bump the release as well and rebuild.

BarryLhm:
omg i forgot it

a/aha/stone.yaml
BarryLhm: > Here you can do `%make_install PREFIX=%(prefix)` and it will install aha and the man pages so you can delete the manual install of the man page. 

i see it but more afraid of the implicitly. what if someday the path of man files or the Makefile would change? (though not so probable)

a/aha/stone.yaml
BarryLhm: lol, running aerynos in qemu where drag-n-drop is not available (as well as kdeconnect because some nat network issue) so typed manually

thanks for careful inspect ^w^

a/aha/stone.yaml
Jaredy899: Fedora and Arch use this approach so if it does change, we can notice at build time. So it will fail, then we can fix. If upstream moves it, that will catch it anyway.

BarryLhm:
all done

Jaredy899 (approved):
LGTM! Thanks!

ermo (approved):

PR #2016, opened 2026-07-17
greetd: Fix the missing service file

irevoire:
**Summary**

greetd was missing its service file when being installed.

**Test Plan**

I rebuilt the package, installed it and systemctl could see the service file.

**Checklist**


irevoire:
After some discussion with @ReillyBrogan I'm closing this PR in favor of https://github.com/AerynOS/recipes/pull/1815/changes#diff-891f9c1e426c4f9a5f610d2e64a58d637b30ee3611f02bcdf982f106cff0d4df

PR #2031, opened 2026-07-20
vesktop: Add at 1.6.5

uxtbd:
**Summary**

Changed the build process from rpm-based to source-based by using pnpm

**Test Plan**

- Login into Vesktop

**Checklist**

**Resolves**:
Resolves #1857 being rpm-based instead of being built from source


v/vesktop/stone.yaml
```diff
@@ -0,0 +1,44 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : vesktop
+version     : "1.6.5"
+release     : 2
+homepage    : https://github.com/Vencord/Vesktop
```
Jaredy899: This should be https://vesktop.dev/

v/vesktop/stone.yaml
```diff
@@ -0,0 +1,44 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : vesktop
+version     : "1.6.5"
+release     : 2
```
Jaredy899: This should be 1

Jaredy899 (changes requested):
And you have 3 commits when there should be one. And as Ermo said, could be a while before this is added.

v/vesktop/stone.yaml
```diff
@@ -0,0 +1,44 @@
...
+homepage    : https://github.com/Vencord/Vesktop
+upstreams   :
+    - https://github.com/Vencord/Vesktop/archive/refs/tags/v1.6.5.tar.gz : 4bb9b5e1acaf17a5f145931008fe4015f9b8c1116b769e1a55a68d5483f238fc
+summary     : Vesktop is a custom Discord App aiming to give you better performance and improve linux support
+description : |
+    Vesktop is a custom Discord App aiming to give you better performance and improve linux support
+license     :
```
Jaredy899, line 13: We should pick one. Arch uses only but the package.json uses or-later. https://github.com/Vencord/Vesktop/blob/8a718e00785f1e17153e49c2d7ffae094e5cecef/package.json#L8
so 
```yaml
license     : GPL-3.0-or-later
```

Jaredy899:
And we should probably put in a patch like `zulip` that disables autoupdates.

uxtbd:
> And we should probably put in a patch like `zulip` that disables autoupdates.

Vesktop doesn't seem to have implemented such feature, no other distro forcefully disables automatic updates for vesktop specifically

uxtbd:
> And you have 3 commits when there should be one. And as Ermo said, could be a while before this is added.

Solved the commits deal, and yeah I have no issue with this not being added right now.

Jaredy899:
Arch strips `publish` from the `package.json`. Patching it like zulip seems sensible.

Jaredy899 (changes requested):

Jaredy899 (dismissed):

NomadicCore:
Hey @uxtbd 

Just checking in. If I am reading things correctly, you had made updates and are waiting a re-review on this recipe?

If that's correct, @Jaredy899, any chance you could take a look?

PR #2036, opened 2026-07-21
shfmt: Add at v3.13.1

irevoire:
# Summary

Add shfmt to format sh files.

# Test Plan

Install and format a few files, both valid and invalid.
Make sure it works in my text editor.
Make sure the manpage is correctly installed and can be opened.

**Checklist**


ermo (changes requested):
- Nit: Please fix up commits to conform to https://github.com/AerynOS/recipes#git-summary-requirements
- Nit: Please ensure that yaml property colons (`:`) are aligned to the same column in the entire stone.yaml recipe file.

irevoire:
> * Nit: Please fix up commits to conform to https://github.com/AerynOS/recipes#git-summary-requirements
> 
>     * Nit: Please ensure that yaml property colons (`:`) are aligned to the same column in the entire stone.yaml recipe file.

Hey, I fixed the colons, but I'm not sure to understand what's wrong with the commit summary?
Or do you means the whole commit message should only be `shfmt: Add at v3.13.1` without the details below?

ermo:
> I'm not sure to understand what's wrong with the commit summary?

image

I think you might simply not have set up our git pre-commit hooks per https://aerynos.dev/packaging/workflow/prerequisites/ ?

ermo (approved):

irevoire:
> I think you might simply not have set up our git pre-commit hooks per https://aerynos.dev/packaging/workflow/prerequisites/ ?

Yeah ok I understood the issue.
I ran the command without thinking too much about it since I never uses git-hooks.
But apparently they're not compatible with `jj`. I'll see if they have an alternative or equivalent to make that work

PR #2040, opened 2026-07-21
freetube: Add at v0.25.1-beta

ey82ysdjia:
FreeTube is an open source desktop YouTube player built with privacy in mind. Use YouTube without advertisements and prevent Google from tracking you with their cookies and JavaScript. Available for Windows (10 and later), Mac (macOS 12 and later) & Linux thanks to Electron.

**Test Plan**
	- Watched something
	- Subscribed to a couple Channels
	- Tried Sponsorblock
	- Changing color to green
	- Used it for a while

**Checklist**


K1ngfish3r:
i approb

empty description though...
license can be on same line
im pretty sure you can use a patch instead of that sed command

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : freetube 
+version     : "0.25.1-beta"
+release     : 1
+homepage    : "https://github.com/FreeTubeApp/FreeTube"
+upstreams   :
+    - https://github.com/FreeTubeApp/FreeTube/archive/refs/tags/v0.25.1-beta.tar.gz : e9e3032264b917e03c76ad38c4ebe9885985e7629dff46712ec40932283d0fb3
+summary     : |
```
Jaredy899: The summary looks like release notes. It should be very short blurb of the package. Like `A private YouTube client`.

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+  This release brings a number of new features along with plenty of bug fixes.
+  Behind the scenes, we're continuing to focus on improving SABR's stability and reliability. Reverse engineering YouTube's custom protocol is a complex and time-consuming process, but it's an investment that will pay off with a more robust experience in future releases.
+description : |
```
Jaredy899: The description should be longer or in this case, it should be there and not empty. (You can probably grab one from the website or github)

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+license     : 
+  - AGPL-3.0
```
Jaredy899: The license can be one line. and from the package.json, it specifies
```yaml
license     : AGPL-3.0-or-later
```

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+builddeps:
```
Jaredy899: The CI is complaining because the builddeps are not in alphabetical order. Also, you can get rid of the soname builddep. It builds fine without it, i.e., it doesn't link against `libcrypt` at build time. We probably don't need any other builddeps other than `pnpm`. `ar` and `tar` are probably not used.

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - binary(pnpm)
+  - binary(ar)
+  - binary(tar)
+  - soname(libcrypt.so.1(x86_64))
+networking: true
+setup       : |
+    # delete package names from config file, 
+    # they say to do this by hand see 
+    # https://docs.freetubeapp.io/development/building-from-source#note-for-linux-users 
+    sed -i "37s/'deb',[[:space:]]*//;37s/'rpm',[[:space:]]*//;37s/,[[:space:]]'pacman'*//;37s/,[[:space:]]'AppImage'*//" _scripts/build.mjs
```
Jaredy899: _Reproducibility_ is key here so we need to patch this rather than a sed. So if something changes upstream, the build will fail. Or else the sed could mess up a future build.

Jaredy899 (changes requested):
The commit message should be `freetube: Add at v0.25.1-beta`. 
Also, it could be a while before we merge it as we are limiting the amount of packages we include. Also, this package may need a conversation around it to include it at all. We can discuss in Zulip if you haven't joined yet.

f/freetube/pkg/freetube.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=FreeTube
+Exec=/usr/bin/freetube
```
Jaredy899: This should probably be `Exec=freetube %u`

f/freetube/pkg/freetube.desktop
```diff
@@ -0,0 +1,10 @@
...
+Terminal=false
+Type=Application
+Icon=freetube
+StartupWMClass=FreeTube
+Comment=A private YouTube client
+MimeType=x-scheme-handler/freetube;
+Categories=Network;
```
Jaredy899: `Categories=AudioVideo;Network;`
See https://specifications.freedesktop.org/menu/latest/category-registry.html

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+    _freetube_install=%(installroot)/$_freetube_root
+    
+    %install_dir $_freetube_install
+    cp -R build/linux-unpacked/* $_freetube_install
+
+    %install_dir %(installroot)%(bindir)
+    ln -svf $_freetube_root/freetube %(installroot)%(bindir)/freetube
+
+    # Desktop file
+    %install_file %(pkgdir)/freetube.desktop %(installroot)/%(datadir)/applications/freetube.desktop
```
Jaredy899: We would need to add the icon. 
`%install_file _icons/icon.svg %(installroot)%(datadir)/icons/hicolor/scalable/apps/freetube.svg`

f/freetube/stone.yaml
```diff
@@ -0,0 +1,41 @@
...
+    pnpm run build
+install     : |
+    _freetube_root=%(libdir)/freetube
+    _freetube_install=%(installroot)/$_freetube_root
+    
+    %install_dir $_freetube_install
+    cp -R build/linux-unpacked/* $_freetube_install
+
+    %install_dir %(installroot)%(bindir)
+    ln -svf $_freetube_root/freetube %(installroot)%(bindir)/freetube
```
Jaredy899: Ideally, Use a relative symlink so /usr/bin/freetube points at the app under /usr/lib without hard-coding an absolute path.
`ln -srv`

Jaredy899 (changes requested):

ey82ysdjia:
@Jaredy899 thanks for taking the time and reviewing and all the notes, i agree with them and made the changes and tested it again

which channel on zulip would be suited to bring this up?

Jaredy899:
You can start a new topic under onboarding.

f/freetube/stone.yaml
Jaredy899: We can probably get rid of the windows and mac stuff. Doesn't really apply to us.

f/freetube/stone.yaml
Jaredy899: We can drop the binaries for `ar` and `tar`. They are not needed here. `Heroic`, `Zulip`, and `Bitwarden` build electron the same way.

f/freetube/stone.yaml
Jaredy899: We should use an actual patch file here. See `zulip-desktop` or `heroic` for examples. so a one line `createTarget(['dir'])` change.

f/freetube/stone.yaml
Jaredy899: We usually have a new line right under this. Just an empty line.

f/freetube/pkg/freetube.desktop
Jaredy899: Ah you know what, I messed this up. The Category should be `AudioVideo;Player;`. AudioVideo and Network are "big" categories in which should only be used once. While "player" is a smaller category we can use here. Plus it prevents duplicate entries in the app menu.

PR #2041, opened 2026-07-21
genfstab: Add at v30

irevoire:
# Summary

Add the genfstab bash script to generate fstab files easily. This is initially a script coming from the arch-install-scripts repo, but it's used in a bunch of other distributions now.

# Test Plan

I installed it, made sure it worked. I also checked the manpage and zsh and bash completions. As far as I can tell, everything works.

**Checklist**


ermo (changes requested):
- This is advertised as v31, but the version field in the recipe is clearly set to "30"? I don't see any patches to justify this discrepancy?
- Nit: Please fix CI builddep sorting nit 
- Nit: Please fix up commits to conform to https://github.com/AerynOS/recipes#git-summary-requirements
- Nit: Please ensure that yaml property colons (`:`) are aligned to the same column in the entire file.

irevoire:
> This is advertised as v31, but the version field in the recipe is clearly set to "30"? I don't see any patches to justify this discrepancy?

That's entirely a mistake on my side; I think it's because initially I only found a fork of this repo and forgot to change the version.

I believe I fixed everything. I'll see how I can set up the hooks so the next reviews are smoother than that

ermo (approved):

PR #2055, opened 2026-07-23
networkmanager: Update to v1.58.0

hphilm:
**Summary**

Changelog: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/raw/1.58.0/NEWS

**Test Plan**

**Checklist**


ermo (commented):
Does this need both newt *and* slang (for consule tui I presume)...?

hphilm:
Seems it also compiles without `slang`. I oriented myself on the Arch PKGBUILD, which added it as a new make dependency.

ermo:
> Seems it also compiles without `slang`. I oriented myself on the Arch PKGBUILD, which added it as a new make dependency.

FTR, if I were given the choice, I'd choose for slang to be enabled. It's got some nice features for TUI apps.

hphilm:
Ya, Reilly should double check. I just wanted to support him on this package. I had to drop `-Dmodify_system=true` as it complained due to security reasons ...

ReillyBrogan (approved):

ReillyBrogan (approved):

PR #2056, opened 2026-07-23
tools(update-cosmic): Add temporary workaround for COSMIC mass update

uxtbd:
**Summary**

Fixes and adds a temporary workaround for the COSMIC mass update

**Test Plan**

The script was used for the 1.4.0 mass update of COSMIC.

**Checklist**

### DIFF
--- a/tools/.shuck.toml
+++ b/tools/.shuck.toml
@@ -1,0 +1,3 @@
+[format]
+indent-style = "space"
+indent-width = 2
--- a/tools/update_cosmic.sh
+++ b/tools/update_cosmic.sh
@@ -5,10 +5,8 @@
 set -e
 set -x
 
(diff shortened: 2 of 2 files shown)


ermo (commented):
Please use 2 spaces of indent, not tabs.

uxtbd:
> Please use 2 spaces of indent, not tabs.

Done, added shucks file for formatting shell scripts within the tools directory, can change it to a new PR if you'd like to focus on that part

hphilm:
Current used scripts to build COSMIC can be found here: https://github.com/hphilm/aerynos-scripts/tree/main/cosmic They also align with the KDE scripts, which need specific build orders. In the end we should provide one script which is able to build multiple packages based on a list to get build orders right ...

PR #2061, opened 2026-07-24
opengamepadui: Update to v0.45.1

hphilm:
**Summary**

- Bugfix release.
- Changelog: https://github.com/ShadowBlip/OpenGamepadUI/releases/tag/v0.45.1

**Test Plan**

- install and check for regressions

**Checklist**


o/opengamepadui/stone.yaml
```diff
@@ -28,7 +31,12 @@ rundeps     :
     - binary(vulkaninfo)
 environment : |
     export PATH="/usr/bin:/usr/sbin"
+    export GODOT="%(bindir)/godot"
+    export GODOT_VERSION="4.6.1"
+    export GODOT_RELEASE="stable"
 setup       : |
+    bsdtar-static --no-same-owner -xf %(sourcedir)/godot.zip --strip-components=0
+    mv Godot_v${GODOT_VERSION}-${GODOT_RELEASE}_linux.x86_64 %(bindir)/godot
```
ermo: There is no guarantee that %(bindir) will stay writable like this, so this solution is not going to fly. Sorry.

Could we make it so that this is instead a vendored godot that is unpacked to / installed in %(libexecdir) by boulder before we make things read-only in the future...?

ermo (commented):
Nvm, the below conflates an install time issue with a pure build time issue (the latter is the problem here).

---
Previous message:
I think we need your input for how to wisely handle truly vendored executables that are not supposed to be called by users, and therefore not supposed to live in the "normal" /usr/bin like shipped moss binaries (i.e. move them out of $PATH)?

In the Linux fileystem standard, such executable binaries/scripts are canonically name-spaced in /usr/libexec/<name>/?

Given we set `%(libexecdir)` to `%(libdir)/%(name)`, perhaps this ought to live in %(libexecdir) during build and be installed to %(installroot)%(libexecdir) ?

o/opengamepadui/stone.yaml
hphilm: Don't know what the issue is here. `godot` is used to build the binaries of opengamepadui and itself won't get installed at all to the system. It is purely a build dependency. We need a certain version to compile opengamepadui. Same solution is used for opengamepadui installer with a much older version of godot. 

For build time the binary of godot can be installed anywhere. We just need to point to it.

Regularly the binary of godot is expected at `/usr/bin`.

hphilm: See also https://github.com/AerynOS/recipes/blob/main/o/opengamepadui-installer/stone.yaml

hphilm: See also: https://github.com/ShadowBlip/OpenGamepadUI/blob/06ffe15fe204bba4c5d3ecfab11983f53d73a348/docker/Dockerfile#L31

ermo: I had missed that it is not installed, so that's that part taken care of.

The fact remains that the buildroot may have its /usr/ tree as exposed to the workdir become read-only in the not-so-distant future, hence flagging it.

hphilm: I'm open for suggestions to place the binary on a different path with write access still ...

ermo: %(workdir)/* will stay writable. Would somewhere in there work for you?

hphilm: Is `%(workdir)/` accessible by the normal PATH structure of the chroot you build in?

ermo: I don't believe it is. You can enter the chroot in any phase by adding a `%break_exit` or a `%break_continue` breakpoint and check the $PATH there (and whatever else of the build state you wish to confirm).

hphilm: Using the `tmp` folder within the boulder `root` environment might be also something ...

ermo: Remember that you can add to $PATH in either a phase or the `environment:` block if you wish to enable it for all phases.

hphilm: We should not add any break stuff or manual editing within this. Normally this is used to use a specific `godot` version within the chroot environment on build time to not need to package multiple versions of `godot`. The binary should be within the build chroot to be able to compile the binaries without a need of any manual steps. How should a CI in a later state handle automatic builds/rebuilds otherwise?

ermo (approved):

PR #2063, opened 2026-07-24
faugus-launcher: Update to 2.0.2

K1ngfish3r:
**Summary**

faugus-launcher: Update to 2.0.2

**Test Plan**

play game

**Checklist**


ermo (approved):
@K1ngfish3r : Do you think we could fruitfully begin to depend on the actual soname(...) providers instead of name providers?

The idea is that we get informed of soname changes (via build failures) re. revdeps basically.

CC: @ReillyBrogan for your thoughts as well.

K1ngfish3r:
> @K1ngfish3r : Do you think we could fruitfully begin to depend on the actual soname(...) providers instead of name providers?
> 
> The idea is that we get informed of soname changes (via build failures) re. revdeps basically.
> 
> CC: @ReillyBrogan for your thoughts as well.

gtk-4, libadawaita are installed by default i believe. if a testing environment can be integrated into boulder this would be best

PR #2065, opened 2026-07-24
kitty: Update to v0.48.1

Jaredy899:
**Summary**
- Wayland fixes
- Release notes can be found [here](https://sw.kovidgoyal.net/kitty/changelog/#detailed-list-of-changes).

**Test Plan**

Pushed commit with `kitty`

**Checklist**


ermo (approved):
@Jaredy899 Maybe we could begin to standardise a 1-3 line summary + a link to release notes as its own item (per my PR edits), so contributors have a guideline?

Jaredy899:
In the commit message or just the PR?

ermo:
> In the commit message or just the PR?

If you do it in the commit, it also shows up in the PR by default, so maybe that's the wisest approach?

What do you see working well here?

Jaredy899:
That's what I usually do. That way the actual commit has the info.

ermo:
> That's what I usually do. That way the actual commit has the info.

Help us lead the way then. =)

PR #2068, opened 2026-07-24
vapoursynth: Update to R78

K1ngfish3r:
**Summary**

- vapoursynth: Update to R78
- zimg updated to current master due to R78 requiring graph engine, will change back to tarball when tag release comes in
- ~~mpv needs `/usr/lib/libvapoursynth-script.so` (unversioned) for vapoursynth support. Currently it isnt being picked up by boulder on mpv side (probably due to the unversioned), so I add it to the base package on vapoursynth side (so user does not have to install the -devel subpackage). This works, but requesting your input~~ 
Fixed, patch mpv so it reads from `/usr/lib/libvapoursynth-script.so.0`

**Test Plan**

Play file on mpv (use vapoursynth filter)

**Checklist**


ermo (approved):
Could we begin adding release note links where appropriate? As a courtesy, I've added the vapoursynth R78 Changelong link for you in this PR.

Also, very good call on adding the run-time selectable avx2 and zn4 (AVX512) .so file here. Let's begin doing this in as many instances as we can for our packages?

PR #2083, opened 2026-07-27
systemd: Update to v260.4

hphilm:
**Summary**

Changelog: https://github.com/systemd/systemd/compare/v260.3...v260.4

**Test Plan**

**Checklist**


ermo:
This appears to have an initrd-related fix according to the commit history (right at the top).

Do we need to bump + rebuild kernels + initramfs(es) in response?

hphilm:
There is another one for the bootloader: https://github.com/systemd/systemd/commit/9e6be8efb52bc0e99f2440df9e49dbd762da9bfd. It depends on how initrd gets generated. On distros like Arch these get generated by the package manager in post-install. In our case it seems to be done during package building as we are stateless. So far Reilly did kernel rebuilds on major systemd releases, not always on point releases.

hphilm:
If I look into my crystal ball and the stable mailing list, new kernel releases may happen this Sunday.. So if `systemd` landed before a bump and just rebuild might not be needed. Unless you plan to have this soon in `unstable` stream for some reason ...

ermo:
I mean, you're the one bumping our kernels at the moment, so your call whether you feel this systemd update warrants a kernel bump in and of itself.

hphilm:
Well, if it is the case to have it 100% working and `systemd` should be used the same exact version used to create the current `initrd` images of what our current kernels ship, then yes. A bump of kernels is logical. However, we had not yet this dependency chain rule. We could have matched `systemd` release with the last `kernels`, as they were released only one day later.

Based on the current speed of kernel upgrades, for my view, it is a very small risk to have a small gap within our `initrd` images on `volatile`, unless you plan an `unstable` stream update with current packages there.

The gap might be even smaller when this lands or even new kernels would be available. Then this needs to been merged before we do kernels.

So, if establish synced `initrd` images in future updates, we should always check if there is a new `systemd` update before we release kernels to the build servers to be compiled ...

ermo:
It probably makes sense to make a note in both the systemd README.md here (if it does not exist, it should be added) and the linux-stable README.md (that the other kernels reference), so that from now on, we make an honest attempt to match up systemd updates that affect initramfs with kernel bumps.

ReillyBrogan (approved):

PR #2097, opened 2026-07-28
Update GitHub workflows

khaga87:
**Summary**

Update our different actions:
actions/checkout from v4 to v7
mshick/add-pr-comment from v2 to v3

**Test Plan**

I'm not sure how to test this without it actually running on recipes. If someone could have a look over it before merging to check my changes?

**Checklist**


ermo (approved):
With our current setup, only one way to find out.

PR #2104, opened 2026-07-28
greetd: Fix greetd not loading /etc/greetd/config.toml

### LINKED ISSUE #2093: greetd ignores /etc/greetd/config.toml
Greetd currently doesn't pick up the standard `config.toml` config file. Only `greetd.conf` or the provided defaults under `/usr` work.

Culprit seems to be here. The `config.toml` branch is never reached.
https://github.com/AerynOS/recipes/blob/d0a06ed980c8067bca2e55141f04a19c28f5a4bd/g/greetd/pkg/0001-Stateless.patch#L27-L33

ChrHorn:
**Summary**

- Fix greetd not loading `/etc/greetd/config.toml`
- Simplify the config loading logic
- Fixed builddeps order (linter complained)

Resolves https://github.com/AerynOS/recipes/issues/2093

**Test Plan**

- Built locally
- Tested with different config names/locations

**Checklist**


ermo:
Re. the patch, this whole set of logic honestly looks like it could use a helper function that then trawls through a prioritised list of config file locations, where the start of the list indicates the highest priority?

ChrHorn:
Fixed the builddeps order, as well as (hopefully) the manifest.

> Re. the patch, this whole set of logic honestly looks like it could use a helper function that then trawls through a prioritised list of config file locations, where the start of the list indicates the highest priority?

That might be a bit overkill imo. The patch now only adds two lines and keeps the existing structure intact. Adding more logic would make it less readable and more brittle as a patch. My Rust skills are also rather limited.

ermo (approved):
Let's give it a go.

PR #2121, opened 2026-07-29
Update: enable xfs_scrub_all.timer for xfs

bhh32:
**Summary**

Enable `xfs_scrub_all.timer` by default so the resilient XFS mount options set by `disks-rs` are periodically scrubbed.

**Test Plan**
  
Build `systemd-preset-aeryn` and verify `xfs_scrub_all.timer is listed as enabled in the preset manifest.

**Checklist**


bhh32:
I rebased this on upstream just before I created the PR. Everything should be ready to merge.

ermo (approved):

PR #2159, opened 2026-08-02
forgejo-cli: Add at v0.6.0

Jaredy899:
**Summary**
- fj is a CLI client for Forgejo, akin to gh, glab, or tea. It handles Forgejo-specific tasks such as issues, pull requests, repositories, organizations, and releases from the command line, working alongside git.

**Test Plan**

Login to codeberg. 

**Checklist**


Jaredy899:
Supersedes #2007

ermo (approved):
Accepting into the repo with the following rationale:

- Acceptable licence
- No other/better option already available
- AerynOS is testing Codeberg repos, which use forgejo
- Package is built from source and tries hard to use native libraries.

PR #2164, opened 2026-08-03
fcitx5-chinese-addons: Add at 5.1.13

K1ngfish3r:
**Summary**

Adds the following packages
- libime
- marisa (bindings for various langs have been split)
- darts (clone by s-yata)
- limonp
- cppjieba
- tclap
- opencc (not sure if should have separate python package or not, left as is)
- fcitx5-lua
- fcitx5-chinese-addons

**Test Plan**

technically not sure what implies a "good" test, i can type... i guess

**Checklist**


ermo:
@lumi-me-not @CookieSource This needs your input I think.

m/marisa/pkg/123.patch
```diff
@@ -0,0 +1,387 @@
...
+ .github/workflows/linux.yml              | 23 +++++++++++++++++++++++
+ bindings/marisa-swig-python3.cxx         | 15 +++++----------
+ bindings/marisa-swig.cxx                 | 15 +++++----------
+ bindings/perl/marisa-swig.cxx            | 15 +++++----------
+ bindings/python/marisa-swig.cxx          | 17 ++++++-----------
+ bindings/python3/marisa-swig-python3.cxx | 15 +++++----------
+ bindings/ruby/marisa-swig.cxx            | 15 +++++----------
+ 7 files changed, 54 insertions(+), 61 deletions(-)
+
+diff --git a/.github/workflows/linux.yml b/.github/workflows/linux.yml
```
lumi-me-not: Why is there a GitHub workflow in the patch?

lumi-me-not (commented):
Admittedly, I don't know enough about the respective codebases to vet the patches very well, but doing a cursory look, it looks good to me. Just a little comment about a GitHub workflow being in the patch, but that shouldn't matter much.

m/marisa/pkg/123.patch
K1ngfish3r: I took the PR as is, could remove it... probably

K1ngfish3r: updated

CookieSource:
I took a peek but I also don't know much about the patches. I am aware the LUA stuff is required for some language dics to work. I should note that outside of Japanese I am not quite sure which ones are the best for other languages. https://wiki.archlinux.org/title/Input_method#List_of_available_input_method_editors I used the Arch page for context. Perhaps it would be wise to ask our Chinese speaking audience what input method works the best for them. I use fcitx5-mozc for Japanese (but asking a native person is still preferable) considering there's multiple choices that work with fcitx5

ermo:
Are we supposed to merge this as-is for more testing?

CookieSource:
I think if typing the characters works lets merge it. Then people can test and we can fix potential issues.

ermo:
> I think if typing the characters works lets merge it. Then people can test and we can fix potential issues.

OK. Then I guess somebody will need to add a video showing that typing the characters works here before we merge.

CookieSource:
@K1ngfish3r Could you provide a quick vid that it works ?

K1ngfish3r:
- removed a stray %break_exit in libime
- fixed cppjieba release-monitoring id leading to another package
- updated all packages to latest

https://github.com/user-attachments/assets/d0e9a39b-37a9-4e50-b3c9-6eb80693f595

f/fcitx5-lua/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
```
Jaredy899, line 4: Needs license info above

Jaredy899 (changes requested):
nit

f/fcitx5-lua/monitoring.yaml
K1ngfish3r: updated

d/darts/manifest.x86_64.jsonc
```diff
@@ -0,0 +1,37 @@
...
+				"binary(darts-benchmark)",
+				"binary(mkdarts)"
+			]
+		},
+		"darts-devel": {
+			"depends": [
+				"darts"
+			],
+			"files": [
+				"/usr/include/darts.h"
```
Jaredy899, line 29: Here with one file in devel. You tell me if we need it or not.

Jaredy899 (commented):
Another

d/darts/manifest.x86_64.jsonc
K1ngfish3r: it's used in opencc i guess

Jaredy899: Sounds good. We can keep it.

Jaredy899 (approved):

PR #2173, opened 2026-08-04
[WIP] system-config-printer: Add at v1.15.18

HarshRelapse:
#### Testing Performed

- [ ] No obvious security or stability concerns or any regressions observed in testing

Additional testing notes:

#### Checklist

- [ ] An approved package request issue exists (link: #[issue number])
- [ ] I am willing to act as the primary maintainer for this package (optional but recommended)

**If you did NOT link to an approved issue,** please explain why in the Additional Context section below.

#### Additional Context
system-config-printer is the main way people add printers to their KDE system, they go to settings -> printers and add their printer. That is the reason for packaging it. Security should also be okay since it is an active project.

#### Screenshots / Output (if applicable)
Bildschirmfoto_20260809_145250

Bildschirmfoto_20260809_145307

```bash
$ sudo moss it system-config-printer
The following package(s) are already installed:

system-config-printer  1.5.18-1
```


K1ngfish3r:
```yaml
builddeps   :
    - binary(xmlto)
    - pkgconfig(cups)
    - pkgconfig(libusb-1.0)
    - python(setuptools)
    - docbook
environment : |
    %export_docbook_catalogs
```

K1ngfish3r:
```yaml
    %reconfigure \
        --sysconfdir=%(datadir) \
        --with-systemdsystemunitdir=%(libdir)/systemd/system \
        --with-udevdir=%(libdir)/udev/rules.d \
        --with-udev-rules \
        --with-cups-serverbin-dir=%(libdir)/cups
```

K1ngfish3r:
I dont believe need to run `%configure` since `%reconfigure` already runs `%configure`

K1ngfish3r:
if using `--sysconfdir=%(datadir) \`, stateless patch not required, since they are both doing the same thing essentially

HarshRelapse:
Deleted reconfigure and didnt build (didn't find ./configure script), so looks like it's needed.

K1ngfish3r:
> Deleted reconfigure and didnt build (didn't find ./configure script), so looks like it's needed.

ie: keep `%reconfigure` and delete the `%configure` inside the build section

HarshRelapse:
I would let %reconfigure stay 1:1(no flags)  since the flags didn't change one bit in the manifest.bin. And yeah configure was useless, thanks for pointing it out again.

K1ngfish3r:
> I would let %reconfigure stay 1:1(no flags) since the flags didn't change one bit in the manifest.bin

ye, it's either stateless patch or build flags

HarshRelapse:
Just tested the recipe on my canon mx 870. Needed to package gutenprint and cups-bjnp for my private repo, but this recipe is tested and I will audit all deps and build logs later.

K1ngfish3r:
can place python(cupshelpers) in a separate package

HarshRelapse:
I am on an roadblock since I encountered an issue that could be more related to cups than this recipe. I cannot seem to get remote printing to work because pdftopdf fails with error 1. system-config-printer worked flawlessly for adding a printer over the kde settings page and even looking at the ink levels and everything else. Since this issue is present even when using the cups webui I belive it's an issue with something else than system-config-printer.

HarshRelapse:
Found the issue, this commit bumping libcupsfilters from 2.1.1 to 2.2.0 https://github.com/aerynOS/recipes/commit/9debc80d1476312288551ff02fe1cd9f34313392, PR dependent on issue https://github.com/aerynOS/recipes/issues/2223

PR #2175, opened 2026-08-05
wf-recorder: ffmpeg rebuild

K1ngfish3r:
**Summary**

wf-recorder: rebuild for ffmpeg 9
~~stolen~~ gratitutious use of `https://github.com/opencv/opencv/pull/29533`

**Test Plan**

install niri, try to pass qsv as a codec arg, fail as hw devices are vaapi only, test vaapi, which works great.

**Checklist**


ReillyBrogan:
...OK but if this causes user-reported issues I'll switch it to ffmpeg 8.1

ReillyBrogan (approved):

PR #2189, opened 2026-08-07
kmscon: Re-enable by default

ReillyBrogan:
I went through all greeters and made sure they conflicted with kmscon on tty1 which should resolve the issues with that.


ermo (approved):
Let's get this in for testing.

hphilm:
Would the following enable `getty` on `tty6` as fallback @ReillyBrogan. How is retriggering of presets done when `epoch` got removed?

```
s/systemd-preset-aeryn/pkg/system/85-override.preset
# Settings in this file are intended to override systemd preset configuration

# Use kmscon instead of getty
disable getty@.service
enable kmsconvt@.service
# ensure we have a fallback getty ootb for when kmscon is not the right tool for the job
enable getty@tty6.service
```

Issue #2288, opened 2026-08-15
build-essential: Add git to package list


NomadicCore:
**Summary**

I am getting set up for packaging on a new system. I have just installed build-essential and am going through the next step on the [prerequisites guide](https://aerynos.dev/packaging/workflow/prerequisites/) which is to set up the helper scripts. This assumes that git is already installed, which, on a fresh install of aerynOS it isn't.

I would suggest that we add the following:
- git
- github-cli
- forgejo-cli

In terms of git and github-cli, these are both needed or very useful for working with aerynOS packaging. forgejo-cli is likely also very useful, especially if we move forward with codeberg.

Jaredy899:
Can we consider this closed?

PR #2290, opened 2026-08-15
build-essential: Update to v0-10

khaga87:
#### Summary

Add the following packages as rundeps:
- boulder
- forgejo-cli
- github-cli
- git

This should make the packaging workflow a little easier as no additional packages should be required as part of the setup process for packaging

#### Testing Performed

Additional testing notes:
I have just set up a new install of aerynOS and run through the packaging guide. These were the packages that I needed to install over and above the build-essential install command.

#### Checklist

- [ ] An approved package update request issue exists (link: #2288)

#### Additional Context

I created an issue for this but it has not actually been approved. Given it's a package update, it's not strictly necessary as per the Package Addition Policy


ermo (approved):
Provisionally approving this on the condition that we keep an eye out for undesirable side effects as reported by users and packagers.

PR #2378, opened 2026-08-23
senpai: Add at v0.5.0

brtz-cs:
#### Testing Performed

Additional testing notes:

#### Checklist

- [x] An approved package request issue exists (fixes : https://github.com/aerynOS/recipes/issues/2187)
- [x] I am willing to act as the primary maintainer for this package (optional but recommended)

**If you did NOT link to an approved issue,** please explain why in the Additional Context section below.

#### Screenshots / Output (if applicable)

senpai_build


s/senpai/monitoring.yaml
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2026 AerynOS Developers
 # SPDX-License-Identifier: MPL-2.0
 releases:
-  id: 142630
-  rss: https://github.com/FiloSottile/age/releases.atom
+  id: 390463
+  rss: https://github.com/delthas/senpai/releases.atom
```
Jaredy899: `rss: https://git.sr.ht/~delthas/senpai/refs/rss.xml`

s/senpai/stone.yaml
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: 2026 AerynOS Developers
+# SPDX-License-Identifier: MPL-2.0
+
+name        : senpai
+version     : "0.5.0"
+release     : 1
+homepage    : https://github.com/delthas/senpai
```
Jaredy899: `homepage    : https://git.sr.ht/~delthas/senpai`

s/senpai/stone.yaml
```diff
@@ -0,0 +1,23 @@
...
+upstreams   :
+    - https://github.com/delthas/senpai/archive/refs/tags/v0.5.0.tar.gz : b1f1e88806fb7d90a3a490d63a3cf2144b7f44392799190570f8e5792a5a706a
```
Jaredy899: https://git.sr.ht/~delthas/senpai/archive/v0.5.0.tar.gz

Jaredy899 (changes requested):
Can you switch these to the actual source? It will need to be rebuilt again. Also, can you post a screenshot of the actual application working?

s/senpai/monitoring.yaml
brtz-cs: I don't know how I missed that, might be because I wrote this at the same time as Age's recipe. Thank you.

s/senpai/stone.yaml
brtz-cs: Regarding this and the other comment about Sourcehut: I knew Sourcehut was the primary source but I considered the Github mirror an easier way to go about it since Boulder can fetch most data for the recipe from Github. However I also realize it's better to get things straight from the source, so I'll fix this.

s/senpai/monitoring.yaml
brtz-cs: Actually, is Github tricking me here? The URL was always right, so why does your comment show the Age one?
github_rename_senpai

brtz-cs: Still, I'll change it to Sourcehut as you said.

Jaredy899: That's weird...

s/senpai/stone.yaml
Jaredy899: Boulder uses Anyita to grab the most recent version from the monitoring.yaml. And Anyita uses the .xml listed above. So when we run `just updates`, it's using data from Anyita to see the most recent version. And therefore `boulder up`, will pull that release.

brtz-cs:
I applied your changes, setting all the sources to Sourcehut. Here's a screenshot of me joining LiberaChat using the client after rebuilding (I only cut out the last part not to show my IP)
senpai_working

Jaredy899 (approved):
LGTM! Friendly tip, when you're ready for another review, you can request a re-review on the top right.

PR #2457, opened 2026-08-30
Global/rebuild

hphilm:


hphilm:
See also https://flit.pypa.io/en/stable/history.html#version-4-0

Jaredy899 (approved):
LGTM

PR #2464, opened 2026-08-31
LLVM: Update to v23.1.0

ReillyBrogan:
- Updated LLVM to v23.1.0
- Enabled lua/tree-sitter support in lldb
- Removed libclc from the llvm build (will be replaced with mesa-libclc in a follow-up PR)
- Added llvm-22
- Rebuild llvm-20 and llvm-21 with LLVM 23


hphilm:
additional kernel patches or updates can be found here: https://github.com/aerynOS/recipes/tree/update/kernels. Next regular kernel bumps expected: `Wed, 02 Sep 2026 13:32:19 +0000`

l/llvm/stone.yaml
```diff
@@ -90,7 +97,8 @@ environment : |
         %patch %(pkgdir)/patches/build/0001-openmp-Don-t-install-static-libarcher.patch
         %patch %(pkgdir)/patches/build/0001-polly-Create-targets-for-headers-cmake.patch
         %patch %(pkgdir)/patches/build/0001-offload-Fix-build-failure-with-icf-all.patch
-        %patch %(pkgdir)/patches/build/0001-Emit-relocations-for-BOLT.patch
+        %patch %(pkgdir)/patches/build/0001-flang-rt-Customize-installdir.patch
+        # %patch %(pkgdir)/patches/build/0001-Emit-relocations-for-BOLT.patch
```
joebonrichie, line 117: Does bolting libclang-cpp/libLLVM actually work with this disabled?

l/llvm/stone.yaml
```diff
@@ -65,19 +81,10 @@ environment : |
...
 
-    EXTRA_LLVM_FLAGS=""
-    if [[ " ${PATH[*]} " =~ "ccache" ]]; then
-        EXTRA_LLVM_FLAGS="-DLLVM_CCACHE_BUILD=ON"
     fi
 
-    # BOLT: -update-debug-sections doesn't yet support SHF_COMPRESSED
-    #       https://github.com/llvm/llvm-project/pull/215835
-    export LDFLAGS="${LDFLAGS} -Wl,--compress-debug-sections=none"
-
```
joebonrichie, line 80: I think this is still needed unless you also want to cherry-pick https://github.com/llvm/llvm-project/pull/215835

l/llvm-22/stone.yaml
```diff
@@ -0,0 +1,652 @@
...
+        # Fix ffmpeg build
+        %patch %(pkgdir)/patches/upstream/0001-SLP-Support-for-tree-throttling-in-SLP-graphs-with-g.patch
+        %patch %(pkgdir)/patches/upstream/0002-X86-vectorizeExtractedCast-don-t-assume-vector-sourc.patch
+        # Fixes rust 1.97 miscompilation
+        %patch %(pkgdir)/patches/upstream/208683.patch
+        # Fix rustc bpf support + prereq
+        %patch %(pkgdir)/patches/upstream/0001-bpf-Support-aggregate-returns-190894.patch
+        %patch %(pkgdir)/patches/upstream/0002-Fixes-for-compile-rust-code-192134.patch
+        # BOLT issues
+        %patch %(pkgdir)/patches/upstream/0001-BOLT-X86-Add-missing-MOV8-instruction-cases-in-X86MC.patch
```
joebonrichie: These bolt patches are superfluous

joebonrichie (changes requested):

l/llvm/stone.yaml
ReillyBrogan: No, the patch probably needs to be rebased but I'll do that on the rebuild (or in the future if it seems like too much work for now)

l/llvm/stone.yaml
ReillyBrogan: 215835.patch

PR #2504, opened 2026-09-03
docker: Update to v29.8.0

Jaredy899:
#### Summary

- Release notes can be found [here](https://docs.docker.com/engine/release-notes/29/).

#### Testing Performed

Additional testing notes:

#### Checklist

- [ ] An approved package update request issue exists (Resolves: #[issue number])


d/docker/stone.yaml
K1ngfish3r: ok

m/moby/stone.yaml
K1ngfish3r: same

K1ngfish3r (changes requested):

d/docker/stone.yaml
Jaredy899: I think we need to be less picky about this. boulder automatically does this.

K1ngfish3r (approved):

PR #2526, opened 2026-09-07
libtirpc: Support for stateless netconfig file

armerpunkt:
#### Summary

* Configure libtirpc so that the `netconfig` file will be installed at `/usr/share/defaults/etc/netconfig`. 
* Add a patch so that libtirpc will check for `/usr/share/defaults/etc/netconfig` after checking for `/etc/netconfig` for stateless functionality.

#### Testing Performed

Additional testing notes:
Recompiled `rpcbind` and `nfs-utils` against `libtirpc` to verify they still compiled and functioned correctly. And tested mounting a nfs share to make sure it mounts correctly.

#### Checklist

- [ ] An approved package update request issue exists

#### Additional Context

Mounting NFS shares would fail due to a missing `/etc/netconfig`. This patch is needed to fix that.


l/libtirpc/stone.yaml
```diff
@@ -18,8 +18,9 @@ environment : |
     # clang 16/17 changes
     export LDFLAGS="${LDFLAGS} -Wl,--undefined-version"
 setup       : |
+    %patch %(pkgdir)/make-netconfig-stateless.patch
     %configure \
-        --disable-static
+        --disable-static --sysconfdir=/usr/share/defaults/etc
 build       : |
```
K1ngfish3r, line 25: turn this into two lines rather than just one

armerpunkt: Done!

K1ngfish3r (dismissed):
i feel like somethings missing that I cant quite put my finger on...
requesting one more approval,

l/libtirpc/stone.yaml
```diff
@@ -18,8 +18,10 @@ environment : |
     # clang 16/17 changes
     export LDFLAGS="${LDFLAGS} -Wl,--undefined-version"
 setup       : |
+    %patch %(pkgdir)/make-netconfig-stateless.patch
     %configure \
-        --disable-static
+        --disable-static \
+        --sysconfdir=/usr/share/defaults/etc
```
Jaredy899: The only nit I can see is this should be `--sysconfdir=%(vendordir)%(sysconfdir)`

Jaredy899 (changes requested):
One thing.

l/libtirpc/stone.yaml
armerpunkt: Done!

Jaredy899 (approved):
LGTM! Thanks!

PR #2572, opened 2026-09-12
linux-stable: Add testing downstream patch for revoking inode watchers

joebonrichie:
#### Summary

Add downstream patch which adds support for `FS_IOC_REVOKE_WATCHES` ioctl.

For discussion only.

#### Checklist

- [ ] This PR links back to an existing issue (Resolves: #[issue number])
- [ ] I have read and conformed to both the [Package Addition Policy](https://github.com/aerynOS/recipes/blob/main/PACKAGING_POLICY.md), and the [Contributing Guidelines](https://aerynos.dev/packaging/). I also confirm that this is my own work or properly attributed and that there are no obvious security or stability concerns or any regressions observed in this testing.
- [ ] No sensitive credentials, API keys, or secrets are included in this PR


hphilm:
Is there a plan to upstream this? Otherwise we would need to maintain this patch downstream for all kernels we may provide for aerynOS ...

joebonrichie:
Don't plan to upstream in it's current state but we could potentially keep it downstream. It does look likely we will need some sort of kernel solution to resolve this issue if we plan to keep using renameat2() w/ RENAME_EXCHANGE as the method to swap in new fs trees.

This is mostly up to generate discussion for the time being

PR #2578, opened 2026-09-14
libxml2: Improve PGO workload & build with 'speed' flags

joebonrichie:
#### Summary

Run the testsuite as part of the PGO profiling to provide a 1.09x speedup when linting shannon's eopkg-index.xml, as well as enable the speed flags for a tiny additional boost.

#### Checklist

- [ ] This PR links back to an existing issue (Resolves: #[issue number])
- [x] I have read and conformed to both the [Package Addition Policy](https://github.com/aerynOS/recipes/blob/main/PACKAGING_POLICY.md), and the [Contributing Guidelines](https://aerynos.dev/packaging/). I also confirm that this is my own work or properly attributed and that there are no obvious security or stability concerns or any regressions observed in this testing.

#### Additional Context

Do not land until a boulder with the fix for https://github.com/aerynOS/os-tools/pull/874 is present on the builders!


ermo (approved):
Looks like there are no soname changes worth worrying about, so no rebuilds to worry about either?

